PR #648, opened 2022-06-23
feat(incentivization): Implemented MVP liquidity gauge distribution

john-connor84:
@NibiruHeisenberg Still missing the hook, tests and cleanup. Just wanted to make sure the logic is correct before implementing the rest and cleaning it up


Unique-Divine:
No longer relevant

PR #1059, opened 2022-11-10
feat(perp): Added query all-positions

onikonychev:
# Description

Allows querying all positions like this:
```
nibid q perp all-positions | jq
```

# Purpose

Allows to track all open positions.

# Blocked by 
- https://github.com/NibiruChain/nibiru/issues/1063
Without pagination it makes no sense.


x/perp/keeper/grpc_query_test.go
```diff
@@ -17,6 +17,50 @@ import (
...
+			Pair:                            common.Pair_BTC_NUSD,
+			Size_:                           sdk.MustNewDecFromStr("1000").MulInt64(i),
+			Margin:                          sdk.MustNewDecFromStr("1000").MulInt64(i),
+			OpenNotional:                    sdk.MustNewDecFromStr("1000").MulInt64(i),
+			LatestCumulativePremiumFraction: sdk.MustNewDecFromStr("1000").MulInt64(i),
+			BlockNumber:                     i,
+		})
+	}
+
+	pos, err := queryServer.QueryAllPositions(sdk.WrapSDKContext(ctx), &types.QueryAllPositionsRequest{})
```
jgimeno, line 58: This test needs kind of an assertion of the positions, is not enough by just checking the error.

Unique-Divine:
Thoughts on making this take a pair argument to be less heavy? You could imagine a case where we have 50+ pairs and this query becomes massive

Unique-Divine:
Closing this in favor of a reminder issue

PR #1082, opened 2022-11-24
feat(vpool): Add governance proposal to edit the swap invariant of one or more vpools

### LINKED ISSUE #1071: feat(perp): define good k values for pools
K is too high in our testnet pool. Figure out what other exchanges are using now, and were using at launch (drift, perp v1 ...)

Unique-Divine:
# Purpose

- Closes #1071

# Change 1 - EditSwapInvariantsProposal

This PR implements a proposal type that maps one or more trading pairs to corresponding multiplers on the swap invariant. The multiplier then changes the virtual reserves in-place without altering the mark price. 
$$
\begin{aligned}
\text{markPrice} = \frac{y}{x} \\
k := x y \\
\text{markPrice}_{\text{new}} = \frac{cy}{cx} = \text{markPrice} \\
k_{\text{new}} = (cx)(cy) = c^2xy = c^2 k
\end{aligned}
$$

The "multiplier" is $c^2$. Notice this implies that a multipying both the base and quote reserves by $c$ results in the new value for $k$ without changing price.

This logic is implemented in `vpool/keeper/pool_state.go` in the `EditSwapInvariant` function. The function and its governance proposal have related unit tests, types tests, marshalling tests, and a CLI test.

# Change 2 - Lower default reserves in localnet.sh


Unique-Divine:
This is a utility for testnet, not live markets.

PR #1097, opened 2022-11-29
feat(perp): Track and expose the net size of a pair with a query

### LINKED ISSUE #1087: feat(perp): Track and expose the net size of a pair with a query

onikonychev:
# Description

Adds new query to `perp` module
```bash
nibid q perp metrics ubtc:unusd | jq
{
  "metrics": {
    "pair": "ubtc:unusd",
    "net_size": "60876.049683584450100149"
  }
}
```

Adds new EndBlocker event `nibiru.perp.v1.MetricsEvent` with the dump of metrics of all vPools.

# Purpose

Adding better trading indicators helps traders making decisions.

- Closes #1087

### DIFF
--- a/x/testutil/cli/network.go
+++ b/x/testutil/cli/network.go
@@ -404,7 +404,7 @@ func (n *Network) LatestHeight() (int64, error) {
 // committed after a given block. If that height is not reached within a timeout,
 // an error is returned. Regardless, the latest height queried is returned.
 func (n *Network) WaitForHeight(h int64) (int64, error) {
-	return n.WaitForHeightWithTimeout(h, 10*time.Second)
+	return n.WaitForHeightWithTimeout(h, 20*time.Second)
 }
 
 // WaitForHeightWithTimeout is the same as WaitForHeight except the caller can


x/perp/keeper/clearing_house.go
```diff
@@ -205,6 +207,22 @@ func (k Keeper) afterPositionUpdate(
...
+
+	iter := k.Positions.Iterate(ctx, collections.PairRange[common.AssetPair, sdk.AccAddress]{})
+	defer iter.Close()
+
+	for ; iter.Valid(); iter.Next() {
+		if pair == iter.Key().K1() {
+			netSize = netSize.Add(iter.Value().Size_)
+		}
+	}
+	k.Metrics.Insert(ctx, pairString, types.Metrics{Pair: pairString, NetSize: netSize})
```
Unique-Divine, line 223: It will be a lot more scalable to compute this when swaps occur on the virtual pool. On calls of `SwapQuoteForBase` and `SwapBaseForQuote`, we can add a small hook/fn to update the global base amount.

Then in the end block hook, we can simply grab the `netSize` from the KV store.

x/perp/keeper/grpc_query_test.go
```diff
@@ -18,6 +18,76 @@ import (
 	"github.com/NibiruChain/nibiru/x/perp/types"
 )
 
+func initAppVpools(
```
Unique-Divine, line 21: 🚀

x/perp/keeper/liquidate.go
```diff
@@ -97,6 +97,8 @@ func (k Keeper) Liquidate(
 		liquidationResponse.FeeToPerpEcosystemFund,
 	)
 
+	k.UpdateMetrics(ctx, pair)
```
Unique-Divine, line 100: Same recommendation as previous comment

Unique-Divine (changes requested):

Unique-Divine (approved):

PR #1102, opened 2022-11-30
refactor(perp)!: replace `CumulativePremiumFractions` array with single value

### LINKED ISSUE #1096: refactor(perp): change `CumulativePremiumFractions` array to single value
Currently, the `CumulativePremiumFractions` is an ever-increasing array of premium fractions. We only ever use the last value in the perps module, but the array is kept around for querying historical values.

If we store the historical values in `heart-monitor` then there's no need to keep the array around, and we can refactor it to a single value.

Depends on https://github.com/NibiruChain/heart-monitor/issues/54.

NibiruHeisenberg:
# Description

- replaces the `CumulativePremiumFractions` array with a single value in the proto definition
- changes the `QueryFundingRates` CLI query with a `QueryCumulativePremiumFraction` query, which returns the latest CPF from the state
- updates tests to use a single value

# Purpose

- Closes #1096

### DIFF
--- a/x/perp/keeper/grpc_query.go
+++ b/x/perp/keeper/grpc_query.go
@@ -2,6 +2,7 @@ package keeper
 
 import (
 	"context"
+	"time"
 
 	"github.com/NibiruChain/collections"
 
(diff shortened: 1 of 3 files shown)


Unique-Divine (approved):

x/perp/client/cli/query.go
```diff
@@ -146,10 +146,10 @@ func CmdQueryPositions() *cobra.Command {
 }
 
 // sample token-pair: btc:nusd
-func CmdQueryFundingRates() *cobra.Command {
+func CmdQueryCumulativePremiumFraction() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "funding-rates [token-pair]",
-		Short: "the cumulative funding payments for a market, up to 48 most recent payments",
+		Short: "the cumulative funding payment for a market",
```
Unique-Divine, line 152:
```suggestion
		Short: "the cumulative funding premium fraction for a market",
```

Unique-Divine (approved):

Unique-Divine:
- Related issue for broken fuzz test: https://github.com/NibiruChain/collections/issues/4

PR #1115, opened 2022-12-11
feat: improve math for single asset join

matthiasmatt:
# Description

@NibiruHeisenberg came up with an easier way to compute the single asset join while at the same time allow for swap fee to be computed: https://www.notion.so/nibiru/Single-Asset-Join-Math-2075178cb9684062b9b65ad23ff14417.

Basically, instead of swapping then joining, we join the pool and then do a single asset join.

This PR migrate the existing single asset join/imbalance join to the new method.

We leverage the uint256 library to do multiplication since we are computing the square of the pool liquidity, which considering the 1e6 precision we use quickly become a very big number (>>int64).

# Purpose

Why is this PR important?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Bug Fixes
 * [#1113](https://github.com/NibiruChain/nibiru/pull/1113) - fix: fix quick simulation issue
 * [#1114](https://github.com/NibiruChain/nibiru/pull/1114) - fix(dex): fix single asset join
+* [#1116](https://github.com/NibiruChain/nibiru/pull/1116) - fix(dex): unfroze pool when LP share supply of 0
 
 ## [v0.16.0](https://github.com/NibiruChain/nibiru/releases/tag/v0.16.0) - Unreleased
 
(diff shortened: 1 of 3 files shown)


x/dex/types/shares.go
```diff
@@ -2,74 +2,68 @@ package types
...
-	var yAmt sdk.Int
-	var xDenom string
+	PRECISION := int64(1_000_000)
 
-	// check who's x and y (x/)
-	if len(tokensIn) == 1 {
-		xAmt = tokensIn[0].Amount
-		xDenom = tokensIn[0].Denom
+	mu := sdk.OneDec().Quo(sdk.OneDec().Sub(pool.PoolParams.SwapFee))
+	sigma := (sdk.OneDec().Add(mu))
```
Unique-Divine, line 30: Let's use semantic variable names instead of μ and σ.

x/dex/types/shares_test.go
```diff
@@ -133,170 +133,67 @@ func TestMaximalSharesFromExactRatioJoin(t *testing.T) {
...
-					Weight: sdk.NewInt(1 << 30),
-				},
-				{
-					Token:  sdk.NewInt64Coin("bbb", 1000),
-					Weight: sdk.NewInt(1 << 30),
-				},
-			},
-			existingShares: 100,
-			tokensIn: sdk.NewCoins(
-				sdk.NewInt64Coin("aaa", 101),
```
Unique-Divine, line 180: I like the new tests that are added. But, we should keep the current test cases (the "aaa" and "bbb" cases) there and update them to have new outputs so that the effect of this PR will be clear just from the side-by-side comparison.

Unique-Divine (changes requested):
very nice

PR #1131, opened 2023-01-04
fix: distribution module account name

NibiruHeisenberg:
# Description

Uses the correct distribution module account name.

# Purpose

Trying to send funds to a non-existent module account causes a panic and a consensus failure.


Unique-Divine:
lol

Unique-Divine (approved):

PR #1139, opened 2023-01-11
feat: set default oracle whitelisted pairs

k-yang:
# Description

Adds assets paired with `NUSD` and `USD` to the oracle params so that the pricefeeder can post prices for it.

# Purpose

We need these oracle prices to enable more perps markets.


k-yang:
Let's merge https://github.com/NibiruChain/nibiru/pull/1141 before this one @jgimeno

k-yang:
> I would put it on the changelog, as it is a change that can affect other tooling expecting those pairs will appear.

Which tools expect those pairs to appear?

Unique-Divine (approved):

PR #1159, opened 2023-01-25
refactor(asset_registry): Use `set[string]` instead of slice `[]string`, since it covers both use cases

Unique-Divine:


Unique-Divine:
These are your changes, so I'm moving it to the original branch

PR #1160, opened 2023-01-26
feat: add generic sets

k-yang:
# Description

Instead of `StringSet`, we can leverage Golang generics to have a set that can handle any "[comparable](https://go.dev/doc/tutorial/generics)" type (i.e. booleans, strings, numbers, etc.).

I also followed the best practices at https://go.dev/blog/package-names and moved the `Set` class to a separate package, so we can use `set.New()` in the code rather than `common.NewSet()` which is more verbose and overloads the `common` package.

# Purpose

Idk, generics seem cool I guess.


Unique-Divine (approved):
Yes, this is the way 🫡

PR #1161, opened 2023-01-27
refactor: migrate simapp tests to use main app

tj327:
# Description

Remove simapp and migrate simulations to use main app.go


jgimeno:
I am not a fan of this issue, this allows to have tests or modules that are still not in the main app

k-yang:
> I am not a fan of this issue, this allows to have tests or modules that are still not in the main app

Which modules did you want to exclude from the main app?

jgimeno:
I see it kind of limiting that we will have only modules wired in the application in all the life of the chain. It looks like it couples only production ready modules when creating new versions or even hotfixes which can make cumberstone to work with different stages of the chain in parallel. 

Given said that I would not get blocked on this branch, I trust the best judgement of engineering team.

k-yang:
> I see it kind of limiting that we will have only modules wired in the application in all the life of the chain. It looks like it couples only production ready modules when creating new versions or even hotfixes which can make cumberstone to work with different stages of the chain in parallel.
> 
> Given said that I would not get blocked on this branch, I trust the best judgement of engineering team.

I think we can leverage long lived development branches to test out new modules or fixes to existing modules, before merging them into release branches.

k-yang (approved):

PR #1162, opened 2023-01-27
fix: multi-platform docker image

k-yang:
# Description

Uses the correct libwasmvm library based on target arch, instead of always using the aarch64 one.

# Purpose

Fixes multi-platform docker images.


jgimeno (approved):

jgimeno:
LGTM

PR #1182, opened 2023-02-07
feat: wasmbinding integration

tj327:
# Description

Cosmwasm binding feature for interacting with perp contract to open andclose perps positions

# Purpose

Cosmwasm binding feature for interacting with perp contract to open andclose perps positions


jgimeno:
This is a good start, I think you will need to use the real simapp as @NibiruHeisenberg removed the old simapp.

tj327:
> This is a good start, I think you will need to use the real simapp as @NibiruHeisenberg removed the old simapp.

Ok

Unique-Divine:
Stale

PR #1195, opened 2023-02-12
feat(perp): multi liquidate

k-yang:
# Description

- wire in multi-liquidate
- add test coverage
- add liquidation failed events
- returns a non-zero status code when all liquidations in the multi liquidate fail

# Purpose

Liquidations are necessary for perps.


proto/perp/v1/tx.proto
```diff
@@ -86,35 +86,32 @@ message MsgAddMarginResponse {
...
-  cosmos.base.v1beta1.Coin fee_to_liquidator = 1 [(gogoproto.nullable) = false];
-  cosmos.base.v1beta1.Coin fee_to_perp_ecosystem_fund = 2 [(gogoproto.nullable) = false];
-}
+// -------------------------- Liquidation --------------------------
 
 message MsgMultiLiquidate {
   string sender = 1;
 
-  message MultiLiquidation {
+  message SingleLiquidation {
```
tj327, line 94: I think its better to use `Liquidation` rather than `SingleLiquidation`

proto/perp/v1/tx.proto
```diff
@@ -86,35 +86,32 @@ message MsgAddMarginResponse {
...
+  repeated SingleLiquidation liquidations = 2;
 }
 
 message MsgMultiLiquidateResponse {
-  message MultiLiquidateResponse {
-    oneof response {
-      string error = 1;
-      MsgLiquidateResponse liquidation = 2;
-    }
+  message SingleLiquidation {
```
tj327, line 106: As this message is used for response, I think `LiquidationResponse` would be more meaningful.

x/perp/client/cli/tx.go
```diff
@@ -29,12 +29,70 @@ func GetTxCmd() *cobra.Command {
...
+		MultiLiquidateCmd(),
 		DonateToEcosystemFundCmd(),
 	)
 
 	return txCmd
 }
 
+func MultiLiquidateCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "multi-liquidate ubtc:unusd:nibi1zaavvzxez0elundtn32qnk9lkm8kmcsz44g7xl ueth:unusd:nibi1zaavvzxez0elundtn32qnk9lkm8kmcsz44g7xl",
```
tj327, line 41: I think this should be changed in better way.
```suggestion
		Use:   "multi-liquidate [Pair1:Trader1] [Pair2:Trader2] ...",
```

Unique-Divine (approved):

PR #1220, opened 2023-03-08
feat: reduce gas fees when posting price

jgimeno:
# Description

This makes that transactions that include a vote or prevote to provide oracle prices is cheaper. Also it enforces that these kind of transactions can only hold one message, so it does no include a message unrelated to provide price data.

- It can contain only one message Vote or Prevote.
- Or maximum 2 if it includes one vote and one prevote.

# Purpose

This is important because right now posting a price is expensive and it does not scale well.


app/ante.go
```diff
@@ -51,8 +51,9 @@ func NewAnteHandler(options AnteHandlerOptions) (sdk.AnteHandler, error) {
 		ante.NewValidateBasicDecorator(),
 		ante.NewTxTimeoutHeightDecorator(),
 		ante.NewValidateMemoDecorator(options.AccountKeeper),
+		feeante.NewPostPriceFixedPriceDecorator(),
 		ante.NewConsumeGasForTxSizeDecorator(options.AccountKeeper),
-		feeante.NewDeductFeeDecorator(options.AccountKeeper, options.BankKeeper, options.FeegrantKeeper), // Replace fee ante from cosmos auth with a custom one.
+		ante.NewDeductFeeDecorator(options.AccountKeeper, options.BankKeeper, options.FeegrantKeeper), // Replace fee ante from cosmos auth with a custom one.
```
Unique-Divine, line 56: Why switch back to the Cosmos SDK one here (`ante`) instead of the custom one (`feeante`)?

app/antedecorators/fee/fixed_gas.go
```diff
@@ -0,0 +1,54 @@
...
+
+	if hasOracleVoteMsg && hasOraclePreVoteMsg {
+		if len(tx.GetMsgs()) > 2 {
+			return ctx, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "a transaction cannot have more than a single oracle vote and prevote message")
+		}
+
+		ctx = ctx.WithGasMeter(types.NewFixedGasMeter(OracleMessageGas))
+	} else if hasOraclePreVoteMsg || hasOracleVoteMsg {
+		if len(tx.GetMsgs()) > 1 {
+			return ctx, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "a transaction that includes an oracle vote or prevote message cannot have more than a single message")
```
Unique-Divine, line 47:
```suggestion
			return ctx, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "a transaction that includes an oracle vote or prevote message cannot have more than those two messages")
```

app/antedecorators/fee/fixed_gas_test.go
```diff
@@ -0,0 +1,231 @@
...
+					Feeder:    addr.String(),
+					Validator: addr.String(),
+				},
+				&types.MsgSend{
+					FromAddress: addr.String(),
+					ToAddress:   addr.String(),
+					Amount:      sdk.NewCoins(sdk.NewInt64Coin(app.BondDenom, 100)),
+				},
+			},
+			expectedGas: 0x151a,
```
Unique-Divine, line 66: For the expected gas, can you use either the `strconv.FormatInt` of `fmt.Sprintf("%x", number)` to make the hexadecimal in the tests readable rather than including directing hexadecimal?

app/ante.go
jgimeno: the custom one was needed for a very specific thing for the old one, this is not needed anymore.

app/antedecorators/fee/fixed_gas.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+
+	"github.com/NibiruChain/nibiru/app/antedecorators/types"
+	oracletypes "github.com/NibiruChain/nibiru/x/oracle/types"
+)
+
+const OracleMessageGas = 500
```
Unique-Divine, line 11: Why 500 specifically?

app/ante.go
Unique-Divine, line 1: A more general refactor point I'm noticing at the end of the review....

Since `app/ante.go` imports from `app/antedecorators/fee` anyway, it seems like we should just have an `app/ante` package that we import from. Then we can easily differentiate the cosmos sdk one as `sdkante` and Nibiru's ante package would be the default `ante`.

Unique-Divine (changes requested):

app/antedecorators/fee/fixed_gas.go
k-yang: We did some back-of-the-envelope calculations. We deduced that this amount of `unibi` would allow oracles to post prices for approximately 3 weeks without running out of gas, using the 100 NIBI obtainable from the faucet.

k-yang (approved):

app/antedecorators/fee/fixed_gas_test.go
jgimeno: absolutely

app/ante.go
jgimeno: good point

Unique-Divine (approved):

PR #1225, opened 2023-03-15
feat: use network in docker compose

jgimeno:
# Description

This is not necessary but it helps as it enforces that a specific network is used and not relay in default account.

# Purpose

Why is this PR important?


jgimeno:
Not needed.

PR #1240, opened 2023-03-28
ci: create workflow to test make proto-gen when the script or .proto files change

Unique-Divine:
This PR adds a workflow to verify that `make proto-gen`  works whenever the `.proto` files change or the proto generation scripts change.

### DIFF
--- a/contrib/scripts/protocgen.sh
+++ b/contrib/scripts/protocgen.sh
@@ -13,7 +13,7 @@ protoc_gen_gocosmos() {
   # get protoc executions
   go get github.com/regen-network/cosmos-proto/protoc-gen-gocosmos@latest 2>/dev/null
 
-  # get cosmos sdk from github
+  # get cosmos sdk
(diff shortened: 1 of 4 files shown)


contrib/scripts/protocgen.sh
```diff
@@ -13,7 +13,7 @@ protoc_gen_gocosmos() {
   # get protoc executions
   go get github.com/regen-network/cosmos-proto/protoc-gen-gocosmos@latest 2>/dev/null
 
-  # get cosmos sdk from github
```
Unique-Divine, line 16: This was a trivial change to make the workflow run

contrib/scripts/protocgen.sh
```diff
@@ -13,7 +13,7 @@ protoc_gen_gocosmos() {
...
+  # get cosmos sdk
```
Unique-Divine, line 16: This was a trivial change added to make the workflow run

jgimeno: XD

PR #1242, opened 2023-03-28
fix: change init pool share supply exponent from 18 to 6

k-yang:
# Description

Changes the pool share supply exponent from 18 (atto) to 6 (micro)

# Purpose

If the initial pool share supply is too large, we may get integer overflow when users add a lot of pool liquidity.


k-yang:
Did some back-of-the-envelope calculations with @Unique-Divine yesterday. It seems that sdk.Int provides us with 2^256-sized integers, which is ~ 10^77. With 10^18 as the base unit for LP shares, it gives us 10^59 orders of magnitude of LP shares, which should be enough for practical purposes.

Also, Osmosis uses 10^18 as their base unit of LP shares, and it seems to work fine in production.

Issue #1236, opened 2023-03-22
feat(perp): add peg multiplier and peg shift logic


matthiasmatt:
We want to introduce the peg multiplier and the peg shift logic to our perp module.

This involves:

- [x] #1232 
- [x] #1233
- [x] #1234
- [x] #1235
- [x] #1241
- [x] #1257

Unique-Divine:
Completed this a while back. Closing as complete

PR #1243, opened 2023-03-28
feat(vpool): sqrt of liquidity depth tracked on pool

### LINKED ISSUE #1241: feat(vpool): store `sqrt_depth` on the vpool struct and use it for computations.

Unique-Divine:
## Summary

- Closes #1241 

| Location | Description |
| --- | --- | 
| x/common | Added functions for computing the square root of `sdk.Dec` and `big.Int` values and corresponding tests. The exact square root is computed using the underlying big integer for the decimal and preserves precision more than root solving.  | 
| proto/vpool | Added `SqrtDepth` as a field of the `Vpool` type. This is part of an iterative change toward using the base reserves, peg multiplier, and bias for pricing swaps. | 
| x/vpool | Integrates changes to the protos, adds validation checks to `Vpool.Validate`, and uses the new square root functions in `EditSwapInvariantProposal` |
| modules like spot, oracle, perp, etc. | Renames or other no-opp refactors |
 
---

## Notes to self (ignore)

- [x] fix types tests
- [x] fix keeper tests
- [x] fix damage dealt to other modules
- [x] fix integration tests
- [x] perform self-review
- [x] write review guide


x/common/dec_test.go
```diff
@@ -0,0 +1,73 @@
...
+	testCases := []struct {
+		dec     sdk.Dec
+		sqrtDec sdk.Dec
+	}{
+		// --------------------------------------------------------------------
+		// Cases: 1 or higher
+		{dec: sdk.NewDec(1), sqrtDec: sdk.NewDec(1)},
+		{dec: sdk.NewDec(4), sqrtDec: sdk.NewDec(2)},
+		{dec: sdk.NewDec(250_000), sqrtDec: sdk.NewDec(500)},
+		{dec: sdk.NewDec(4_819_136_400), sqrtDec: sdk.NewDec(69_420)},
```
Unique-Divine, line 45: Add a few large number cases to show that 10^18 precision is conserved.

x/common/dec.go
```diff
@@ -0,0 +1,59 @@
...
+	var panicErr error = TryCatch(func() {
+		sqrtDec = MustSqrtDec(dec)
+	})()
+	return sqrtDec, panicErr
+}
+
+// SqrtBigInt returns the square root of the input.
+//   - NOTE: SqrtBigInt panics if it is called on a negative number because it uses
+//     the `big.Int.Sqrt` from the "math/big" package.
+func SqrtBigInt(i *big.Int) *big.Int {
```
Unique-Divine, line 31: This should be `MustSqrtBigInt`, and `SqrtBigInt` should have an error in the return

x/common/dec.go
```diff
@@ -0,0 +1,59 @@
...
+	PRECISION = 18
+)
+
+var (
+	PRECISION_MULT = calcPrecisionMultiplier(0)
+	PRECISION_SQRT = int64(PRECISION / 2)
+	tenInt         = big.NewInt(10)
+)
+
+// calculate the precision multiplier
```
Unique-Divine, line 51: Although the Cosmos SDK doesn't have docs on this function, ours should.

jgimeno:
I lack some context on why this change. Is there any doc?

Unique-Divine:
@jgimeno Missing fields from the AMM struct here: https://github.com/NibiruChain/think-tank/blob/794b28c3951a7605b8832595c03dbc82b46d2070/drift-v1/programs/clearing_house/src/state/market.rs#L90-L107

This is one of the pieces needed to close 
- #1236 
- #1241

proto/vpool/v1/state.proto
```diff
@@ -48,7 +50,27 @@ message Vpool {
     (gogoproto.nullable) = false
   ];
 
-  VpoolConfig config = 4 [
+  // The square root of the liquidity depth. Liquidity depth is the product of
```
jgimeno, line 53: This change enforces a restart of the network, if this is not expected I would at least use a new number for new fields as protocol buffers mention.

https://learn.microsoft.com/en-us/dotnet/architecture/grpc-for-wcf-developers/protobuf-messages#field-numbers

jgimeno: Yes, ideally no same field should change number. Even it is recommended when deprecating a field to leave that number set up so it cannot be taken anymore.

In this case config should keep the 4.

x/common/constants.go
```diff
@@ -3,5 +3,5 @@ package common
 const (
 	TreasuryPoolModuleAccount = "treasury_pool"
 	// Precision for int representation in sdk.Int objects
-	Precision = int64(1_000_000)
+	MICRO = int64(1_000_000)
```
k-yang, line 6: `MICRO` or `MEGA`? Also what's wrong with `Precision`?

In my mind, 
micro ==> multiply by 10^-6
mega ==> multiply by 10^6

Unique-Divine: If you have 2 dollars, you have 2 million micro dollars. Micro is multiplication.

`Precision` is vague, and someone wouldn't know what it means by looking at it. In our case, we're not really using this value as a precision. It's just a simple multiplier into micro units.

k-yang: Yes, `TO_MICRO` or `AS_MICRO` would be more clear.

proto/vpool/v1/state.proto
```diff
@@ -51,6 +53,26 @@ message Vpool {
...
+  // the reserves.
+  string sqrt_depth = 5 [
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec",
+    (gogoproto.nullable) = false
+  ];
+
+  // Bias refers to the net long-short bias of the market. The bias of a pool 
+  // is equivalent to the sum of all base reserve changes since its creation.
+  // string bias = 5 [
```
k-yang, line 66: Do we want to include these commented-out fields in this PR?

x/vpool/keeper/pool_state_test.go
```diff
@@ -44,8 +43,9 @@ func TestEditPoolConfig(t *testing.T) {
 	pair := asset.Registry.Pair(denoms.BTC, denoms.NUSD)
 	vpoolStart := types.Vpool{
 		Pair:              pair,
-		QuoteAssetReserve: sdk.NewDec(10 * common.Precision),
-		BaseAssetReserve:  sdk.NewDec(5 * common.Precision),
+		QuoteAssetReserve: sdk.NewDec(10 * common.TO_MICRO),
+		BaseAssetReserve:  sdk.NewDec(5 * common.TO_MICRO),
+		SqrtDepth:         common.MustSqrtDec(sdk.NewDec(5 * 10 * common.TO_MICRO * common.TO_MICRO)),
```
k-yang, line 48: In this case, I would put in the literal number we expect rather than rely on a utility function to calculate it for us, because the utility function is also used in the class's code path to calculate this value. If something changes in the utility function, this test case would still pass even though the final value isn't what we expect for this vpool entity.

k-yang (commented):
A couple of nits, but overall LGTM

k-yang (approved):

x/vpool/keeper/pool_state_test.go
Unique-Divine: I added a simplified constructor for this. I'll add a ticket to convert over most of the `vpooltypes.Vpool` initializations to use `vpooltypes.NewVpool(types.ArgsNewVpool(...))` to accomplish this more directly. 
- https://github.com/NibiruChain/nibiru/issues/1253
I started working on this in the vpool module, but the PR was becoming large. 

Context: A Slack discussion

image

PR #1271, opened 2023-04-08
refactor(perp)!: vpool → perp/amm #2 | imports and renames

Unique-Divine:
## Removes "vpool" from the codebase

- This is a pure refactor
- Completes changes that follow #1267


k-yang (approved):

jgimeno:
This is not what was expected, there is no benefit on this move than having it as before.

Unique-Divine:
It's incomplete because the keepers and KV stores aren't yet combined. The move is done in pieces to make it more manageable

PR #1281, opened 2023-04-14
feat(perp): use peg multiplier to compute prices

matthiasmatt:
# Description

Use peg and bias to compute the price of a pool.

# Purpose

New logic that can be repegged


proto/perp/amm/v1/state.proto
```diff
@@ -145,7 +145,19 @@ message ReserveSnapshot {
   ];
 
   // milliseconds since unix epoch
-  int64 timestamp_ms = 3;
```
jgimeno, line 148: Do you know why there is this change in number? In theory this is not recommended.

x/perp/amm/abci_test.go
```diff
@@ -31,21 +31,21 @@ func TestSnapshotUpdates(t *testing.T) {
...
-		*types.DefaultMarketConfig().
-			WithTradeLimitRatio(sdk.OneDec()).
-			WithFluctuationLimitRatio(sdk.OneDec()),
-		sdk.ZeroDec(),
-		sdk.OneDec(),
+		/* ctx */ ctx,
+		/* pair */ asset.Registry.Pair(denoms.BTC, denoms.NUSD),
+		/* quoteReserve */ sdk.NewDec(1_000),
+		/* baseReserve */ sdk.NewDec(1_000),
+		/* config */ *types.DefaultMarketConfig().WithTradeLimitRatio(sdk.OneDec()).WithFluctuationLimitRatio(sdk.OneDec()),
```
jgimeno, line 38: I would set defaultmarketconfig as type, not pointer, to avoid this casting.

jgimeno (approved):

x/perp/amm/abci_test.go
Unique-Divine: Golang doesn't have the default args feature, you have have to use pointer if you want to see default values inside a constructor

x/perp/amm/keeper/bias_test.go
```diff
@@ -55,7 +55,7 @@ func TestBiasChangeOnMarket(t *testing.T) {
 			).
 			Then(
 				ammassertion.MarketShouldBeEqual(pairBtcUsdc,
-					ammassertion.Market_BiasShouldBeEqualTo(sdk.MustNewDecFromStr("9999.999900000001000000")), // Bias equal to PositionSize
+					ammassertion.Market_BiasShouldBeEqualTo(sdk.MustNewDecFromStr("-9999.999900000001000000")), // Bias equal to PositionSize
 				),
```
Unique-Divine, line 59: These test cases are opposite. The trader opened long, so the bias should be positive. Semantically, we'd be saying the market has a long (positive) bias.

x/perp/amm/keeper/bias_test.go
```diff
@@ -75,7 +75,7 @@ func TestBiasChangeOnMarket(t *testing.T) {
 			).
 			Then(
 				ammassertion.MarketShouldBeEqual(pairBtcUsdc,
-					ammassertion.Market_BiasShouldBeEqualTo(sdk.MustNewDecFromStr("19999.999600000008000000")), // Bias equal to PositionSize
```
Unique-Divine, line 78: same (opposite)

x/perp/amm/keeper/bias_test.go
```diff
@@ -92,7 +92,7 @@ func TestBiasChangeOnMarket(t *testing.T) {
 			).
 			Then(
 				ammassertion.MarketShouldBeEqual(pairBtcUsdc,
-					ammassertion.Market_BiasShouldBeEqualTo(sdk.MustNewDecFromStr("-10000.000100000001000000")), // Bias equal to PositionSize
+					ammassertion.Market_BiasShouldBeEqualTo(sdk.MustNewDecFromStr("10000.000100000001000000")), // Bias equal to PositionSize
 				),
 				PositionShouldBeEqual(alice, pairBtcUsdc, Position_PositionSizeShouldBeEqualTo(sdk.MustNewDecFromStr("-10000.000100000001000000"))),
```
Unique-Divine, line 97: same (opposite)

x/perp/amm/keeper/bias_test.go
```diff
@@ -112,7 +112,7 @@ func TestBiasChangeOnMarket(t *testing.T) {
 			).
 			Then(
 				ammassertion.MarketShouldBeEqual(pairBtcUsdc,
-					ammassertion.Market_BiasShouldBeEqualTo(sdk.MustNewDecFromStr("-20000.000400000008000000")), // Bias equal to PositionSize
+					ammassertion.Market_BiasShouldBeEqualTo(sdk.MustNewDecFromStr("20000.000400000008000000")), // Bias equal to PositionSize
 				),
```
Unique-Divine, line 116: same (opposite)

x/perp/amm/keeper/bias_test.go
```diff
@@ -150,7 +150,7 @@ func TestBiasChangeOnMarket(t *testing.T) {
 			).
 			Then(
 				ammassertion.MarketShouldBeEqual(pairBtcUsdc,
-					ammassertion.Market_BiasShouldBeEqualTo(sdk.MustNewDecFromStr("8999.999919000000729000")), // Bias equal to PositionSize
+					ammassertion.Market_BiasShouldBeEqualTo(sdk.MustNewDecFromStr("-8999.999919000000729000")), // Bias equal to PositionSize
 				),
```
Unique-Divine, line 154: same (opposite)

x/perp/amm/keeper/prices.go
```diff
@@ -59,37 +59,12 @@ func (k Keeper) GetBaseAssetPrice(
...
-	pool, err := k.Pools.Get(ctx, pair)
+	quoteAmount, err = amm.GetQuoteAmountByBaseAmount(baseAssetAmount.MulInt64(dir.ToMultiplier()))
 	if err != nil {
-		return sdk.ZeroDec(), err
+		return
 	}
-
-	dirMult := dir.ToMultiplier()
-	return pool.GetBaseAmountByQuoteAmount(quoteAmountAbs.MulInt64(dirMult))
+	quoteAmount = quoteAmount.Mul(amm.PegMultiplier)
```
Unique-Divine, line 66: It looks like we were supposed to multiply by peg, then divide base. Is the input arg in assets or reserves?

Unique-Divine (changes requested):
Review still WIP, coming back shortly

proto/perp/amm/v1/state.proto
k-yang: 3 and 5 are missing field numbers.

x/perp/amm/abci_test.go
k-yang: It should be possible to pass a struct instead of a pointer to a struct.

k-yang (approved):

PR #1282, opened 2023-04-16
feat(inflation)!: Add inflation module

k-yang:
# Description

Adds the x/inflation module, heavily inspired [from Evmos](https://github.com/evmos/evmos/tree/main/x/inflation).

Most of the LOC are from the auto-generated proto files.

# Purpose

Codifies our tokenomics emission schedule for 8 years.


x/inflation/types/params_test.go
k-yang, line 1: This file just tests valid/invalid params.

x/inflation/types/inflation_calculation_test.go
k-yang, line 1: This file is the important one, as it tests the actual inflation per epoch.

PR #1296, opened 2023-04-28
refactor(perp)!: perp data model updates

k-yang:
# Description

Builds upon #1291 with additional tweaks to the perps state protos.

# Purpose

Clean state protos for a clean domain model.


x/perp/types/v2/msgs.go
k-yang, line 1: Necessary for serializing perp msg protos.

x/perp/types/v2/codec.go
k-yang, line 1: Forgot to add codec in #1291.

proto/perp/v1/state.proto
```diff
@@ -145,89 +145,6 @@ message PrepaidBadDebt {
   ];
 }
 
-message PositionResp {
```
k-yang, line 148: Turned `PositionResp` and `LiquidateResp` into Golang structs since they weren't being serialized or passed around anywhere.

Unique-Divine (approved):

PR #1307, opened 2023-05-02
feat(sudo): Implements the x/sudo module

Unique-Divine:
# Summary

- feat(sudo): implement all business logic with full test coverage
- test: add integration tests of the sudo contracts in x/wasm
- test: adds full test coverage in x/sudo.

# Description

The `x/sudo` provides a simple way to manage, verify, and run smart contracts
with elevated permissions, mimicking the functionality of a Unix-based sudoers
file.

Most of the `sudo` module fits in one file: sudo.go.

In the context of Unix-based operating systems, such as Linux and macOS,
"sudoers" refers to a configuration file that defines which users have the
privilege to execute commands with elevated (superuser) permissions using the
"sudo" command. The superuser, also known as the root user, has full
administrative access to the system, allowing them to perform tasks that regular
users cannot do.

### State

The module state is a simple data structure. 

```go
type Sudoers struct {
	Root      string          `json:"root"`
	Contracts set.Set[string] `json:"contracts"`
}
```

The `Sudoers` state is configurable in the genesis and can only be modified afterward by the "root" account (or one of the sudo contracts if we implement special bindings for `x/sudo` in the future).

> **Note**
>  Other modules that wish to use module bindings permissioned to specific CosmWasm smart contracts should implement keeper-only functions that execute via the `CustomWasmExecutor` in the `x/wasm/binding` package.

# Purpose

See [Nibi-Perps Admin ADR](https://www.notion.so/nibiru/Nibi-Perps-Admin-ADR-ad38991fffd34e7798618731be0fa922?pvs=4).

# Next Steps

- [ ] Make sure events emit as expected in `x/wasm`.
- [ ] Add a CLI transaction command for `x/sudo`.


jgimeno (approved):
It looks great

x/sudo/sudo.go
```diff
@@ -0,0 +1,236 @@
...
+func NewKeeper(
+	cdc codec.BinaryCodec,
+	storeKey sdk.StoreKey,
+) Keeper {
+	return Keeper{
+		Sudoers: collections.NewItem(storeKey, 1, SudoersValueEncoder(cdc)),
+	}
+}
+
+var ROOT_ACTIONS = pb.ROOT_ACTIONS
```
jgimeno, line 31: this is not used I think

x/sudo/sudo.go
```diff
@@ -0,0 +1,236 @@
...
+			return nil, sdkerrors.Wrap(sdkerrors.ErrUnknownRequest, errMsg)
+		}
+	}
+}
+
+// Ensure the interface is properly implemented at compile time
+var _ pb.MsgServer = Keeper{}
+
+func (k Keeper) EditSudoers(
+	goCtx context.Context, msg *pb.MsgEditSudoers,
```
jgimeno, line 56: I would validate the message outside of this function, and the default should return an error.

x/sudo/sudo.go
```diff
@@ -0,0 +1,236 @@
...
+		return
+	}
+
+	// Read state
+	ctx := sdk.UnwrapSDKContext(goCtx)
+	pbSudoersBefore, err := k.Sudoers.Get(ctx)
+	if err != nil {
+		return
+	}
+	sudoersBefore := Sudoers{}.FromPbSudoers(pbSudoersBefore)
```
jgimeno, line 154: you can go wild and create a proto encoder for the Set type so you dot have to use this converters.

x/sudo/sudo.go
Unique-Divine: Good idea, updated!

PR #1312, opened 2023-05-03
feat: wire edit swap invariant to the contracts

matthiasmatt:
# Description

What does this PR do?

# Purpose

Why is this PR important?


x/perp/keeper/amm.go
```diff
@@ -27,30 +27,91 @@ func (k Keeper) EditPoolPegMultiplier(
...
+	err = k.PerpAmmKeeper.EditSwapInvariant(ctx, pair, multiplier)
+	if err != nil {
+		return
+	}
+
+	// Get the pool
+	pool, err = k.PerpAmmKeeper.GetPool(ctx, pair)
+	if nil != err {
+		return
+	}
```
Unique-Divine, line 81: It looks like this is an unnecessary duplicate read since you already have the pool value in a variable

Unique-Divine (changes requested):
Nice work. I added an updated smart contract and setup function for it to the `x/wasm` tests. I only saw 1 change

Unique-Divine (approved):

Unique-Divine (approved):

PR #1315, opened 2023-05-04
feat: oracle rewards distribution every week

jgimeno:
# Description

What does this PR do?

# Purpose

Why is this PR important?


proto/oracle/v1/genesis.proto
```diff
@@ -26,7 +26,7 @@ message GenesisState {
         "github.com/NibiruChain/nibiru/x/common/asset.Pair",
     (gogoproto.nullable) = false
   ];
-  repeated PairReward pair_rewards = 8 [ (gogoproto.nullable) = false ];
+  repeated Reward pair_rewards = 8 [ (gogoproto.nullable) = false ];
```
k-yang, line 29: I would rename the field to just `rewards`.

x/epochs/keeper/keeper.go
```diff
@@ -3,25 +3,29 @@ package keeper
...
+
+	Epochs collections.Map[string, types.EpochInfo]
+}
 
 func NewKeeper(cdc codec.Codec, storeKey sdk.StoreKey) Keeper {
 	return Keeper{
 		cdc:      cdc,
 		storeKey: storeKey,
+
+		Epochs: collections.NewMap[string, types.EpochInfo](storeKey, 1, collections.StringKeyEncoder, collections.ProtoValueEncoder[types.EpochInfo](cdc)),
```
k-yang, line 28: ❤️

x/epochs/keeper/epoch.go
```diff
@@ -59,35 +60,27 @@ func (k Keeper) AddEpochInfo(ctx sdk.Context, epoch types.EpochInfo) error {
 
 // UpsertEpochInfo inserts the epoch if does not exist, and overwrites it if it does.
 func (k Keeper) UpsertEpochInfo(ctx sdk.Context, epoch types.EpochInfo) {
-	store := ctx.KVStore(k.storeKey)
-	value, err := proto.Marshal(&epoch)
-	if err != nil {
-		panic(err)
-	}
-	store.Set(append(types.KeyPrefixEpoch, []byte(epoch.Identifier)...), value)
+	k.Epochs.Insert(ctx, epoch.Identifier, epoch)
```
k-yang, line 63: Are these methods of any value? In my opinion, any function whose implementation is one line long can be replaced with executing that actual line.

x/oracle/keeper/hooks.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+func (h Hooks) AfterEpochEnd(ctx sdk.Context, epochIdentifier string, _ uint64) {
+	if epochIdentifier == types.WeekEpochID {
+		account := h.accountKeeper.GetModuleAccount(ctx, perptypes.FeePoolModuleAccount)
+		totalValidatorFees, totalRest := sdk.Coins{}, sdk.Coins{}
+
+		balances := h.bankKeeper.GetAllBalances(ctx, account.GetAddress())
+		for _, balance := range balances {
+			validatorFees := balance.Amount.ToDec().Mul(sdk.MustNewDecFromStr("0.05")).TruncateInt()
```
k-yang, line 41: Let's avoid a hard-coded percentage value here and make it a Param? I think you were going to refactor the x/oracle Params anyways.

x/oracle/keeper/keeper.go
```diff
@@ -41,17 +41,17 @@ type Keeper struct {
 	// PriceSnapshots maps types.PriceSnapshot to the asset.Pair of the snapshot and the creation timestamp as keys.Uint64Key.
 	PriceSnapshots   collections.Map[collections.Pair[asset.Pair, time.Time], types.PriceSnapshot]
 	WhitelistedPairs collections.KeySet[asset.Pair]
-	PairRewards      collections.IndexedMap[uint64, types.PairReward, PairRewardsIndexes]
+	PairRewards      collections.Map[uint64, types.Reward]
```
k-yang, line 44:
```suggestion
	Rewards      collections.Map[uint64, types.Reward]
```

x/oracle/keeper/keeper.go
```diff
@@ -41,17 +41,17 @@ type Keeper struct {
...
 	PairRewardsID    collections.Sequence
 }
 
 type PairRewardsIndexes struct {
 	// RewardsByPair is the index that maps rewards associated with specific pairs.
-	RewardsByPair collections.MultiIndex[asset.Pair, uint64, types.PairReward]
+	RewardsByPair collections.MultiIndex[asset.Pair, uint64, types.Reward]
```
k-yang, line 50: Does a `RewardsByPair` index even make sense anymore? I think we can get rid of this struct.

x/epochs/keeper/epoch.go
jgimeno: they have a value in some specific cases, probably not on inserting, but when we want to define expected_keepers interface.

Imagine that you want to get the Epoch with ID x, you cannot define the Epochs type in an interface. 

there is a thin line on when to do it and when not.

The ideal would be that a keeper only have the exported methods it is allowed to be called from other modules and they do not break your data in unexpected ways.

k-yang: As it stands right now, other keepers can directly access the collection fields of this keeper right?

proto/oracle/v1/oracle.proto
```diff
@@ -122,16 +122,9 @@ message ExchangeRateTuple {
...
-  // pair defines the pair for which we incentivize validator to provide prices
-  // for.
-  string pair = 1 [
-    (gogoproto.customtype) =
-        "github.com/NibiruChain/nibiru/x/common/asset.Pair",
-    (gogoproto.nullable) = false
-  ];
+message Rewards {
   // id uniquely identifies the rewards instance of the pair
   uint64 id = 2;
```
k-yang, line 129: Maybe shift the field numbers back to 1?

x/epochs/keeper/epoch.go
jgimeno: depends if you use the pattern of expected keepers or not

x/oracle/keeper/keeper.go
jgimeno: Good point, I removed the logic but I totally forgot the hanging struct.

proto/oracle/v1/oracle.proto
jgimeno: it was out of proto habit! :D

proto/oracle/v1/genesis.proto
```diff
@@ -26,7 +26,7 @@ message GenesisState {
         "github.com/NibiruChain/nibiru/x/common/asset.Pair",
     (gogoproto.nullable) = false
   ];
-  repeated PairReward pair_rewards = 8 [ (gogoproto.nullable) = false ];
+  repeated Rewards pair_rewards = 8 [ (gogoproto.nullable) = false ];
```
k-yang, line 29:
```suggestion
  repeated Rewards rewards = 8 [ (gogoproto.nullable) = false ];
```

k-yang (approved):

PR #1317, opened 2023-05-05
feat(sudo): CLI fns + tests. Also, switches testutil to use secp256k1

### LINKED ISSUE #1316: feat(testutil): key generation should support both ed25519 and  secp256k1.
## Context 

All of our key generation util functions in `x/common/testutil` are generated from ed25519. I found out when trying to import one of these accounts to the validator node's `clientCtx.Keyring` that support was never added on the binary for ed25519 keys.

- [Background](#background)
- [Reproducibility](#reproducibility) 
- [Desired Solution](#desired-solution)

## Background

It seems the feature was de-prioritized and never implemented according to the comments in the SDK repo:

> [zmanian](https://github.com/cosmos/cosmos-sdk/issues/1861#issuecomment-410535591): "Disagree on Prelaunch. All fundraiser keys are secp keys. No one can even really use a new keypair until transfers are enabled."

> [ValarDragon](https://github.com/cosmos/cosmos-sdk/issues/1861#issuecomment-410537934): "That sounds good to me! This is definitely something that will be easier to support later. (Doesn't even involve changing the version) I think I'll just close this, and we can look into ed25519 again post launch / after hd derivation is suitably standardized."

- https://github.com/cosmos/cosmos-sdk/issues/1861
- https://github.com/cosmos/cosmos-sdk/issues/9227

## Reproducibility

 This was some of the code I wrote that made the error apparent:

```go
import (
  // ...
  "github.com/cosmos/cosmos-sdk/crypto"
  cryptotypes "github.com/cosmos/cosmos-sdk/crypto/types"
)

type Account struct {
	privKey    cryptotypes.PrivKey
	addr       sdk.AccAddress
	passphrase string
}

func (s *IntegrationSuite) AddRootToKeyring(root Account) {
	s.T().Log("add the x/sudo root account to the clientCtx.Keyring")
	// Encrypt the x/sudo root account's private key to get its "armor"
	passphrase := root.passphrase
	privKey := root.privKey
	armor := crypto.EncryptArmorPrivKey(privKey, passphrase, privKey.Type())
	// Import this account to the keyring
	val := s.network.Validators[0]
	s.NoError(
		val.ClientCtx.Keyring.ImportPrivKey("root", armor, passphrase),
	)
}
```

The `root` account was created using the `PrivKeyAddresPairs` function. The semantically relevant part of that function is:
```go
func PrivKeyAddressPairs(n int) (keys []cryptotypes.PrivKey, addrs []sdk.AccAddress) {
    // ... {
    keys[i] = ed25519.GenPrivKeyFromSecret(secret)
    addrs[i] = sdk.AccAddress(keys[i].PubKey().Address())
  }
  return
}
```

I executed a transaction with the `root` as the sender and got the following output. 
```json
{
  "height": "0",
  "txhash": "58DBC804E118DC1D4BED13708BD2F3A15B41A61FF1828A40442DDA0BD1513DCA",
  "codespace": "sdk",
  "code": 8,
  "data": "",
  "raw_log": "ED25519 public keys are unsupported: invalid pubkey",
  "logs": [],
  "info": "",
  "gas_wanted": "200000",
  "gas_used": "53347",
  "tx": null,
  "timestamp": "",
  "events": []
}
```

## Desired Solution

Both the `testutil.AccAddress` and `testutil.PrivKeyAddressPairs` functions in `x/common/testutil` should generate keys with secp256k1 instead of ed25519. We can replace or rename the current versions of these functions.

Unique-Divine:
- Implements and tests CLI functions for the x/sudo module.
- Updates `common/testutil` to use secp256k1 instead of ed25519. 
- Closes #1316


x/common/testutil/sample.go
```diff
@@ -13,16 +15,52 @@ import (
...
+		}
+		keys[i] = secp256k1.GenPrivKeyFromSecret(secret)
+		addrs[i] = sdk.AccAddress(keys[i].PubKey().Address())
+	}
+	return
+}
+
+// Ed25519 functions as a wrapper class implementing the "AccAddress" and
+// "PrivKeyAddressPairs" methods using ed25519.
+var Ed25519 ed25519Algo = ed25519Algo{}
```
jgimeno, line 52: why we add this?

jgimeno: Just saw your description, but adding this without having really a full support for creating accounts using Ed25519 is out of the scope of the PR.

Unique-Divine: I only added that to avoid deleting it entirely. I'm fine with removing it too

Unique-Divine: 5eea4a77e2276696b13a7d75fe1425693bfe0d23

PR #1318, opened 2023-05-07
feat(perp): add PerpKeeperV2 Calc and Twap methods

k-yang:
# Description

## Calc Methods
- PositionNotionalSpot
- PositionNotionalTWAP
- UnrealizedPnl
- MarginRatio
- FundingPayment

## Twap Methods
- CalcTwap
- getPriceWithSnapshot

# Purpose

Refactors the PerpKeeper to use new state


x/perp/integration/action/v2/market.go
```diff
@@ -0,0 +1,79 @@
...
+
+	app.PerpKeeperV2.ReserveSnapshots.Insert(ctx, collections.Join(c.AMM.Pair, ctx.BlockTime()), v2types.ReserveSnapshot{
+		Amm:         c.AMM,
+		TimestampMs: ctx.BlockTime().UnixMilli(),
+	})
+
+	return ctx, nil, true
+}
+
+// CreateCustomMarket creates a market with custom parameters
```
jgimeno, line 37: Love it

x/perp/integration/action/v2/snapshot.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	if !twap.Equal(c.expectedTwap) {
+		return ctx, fmt.Errorf("invalid twap, expected %s, received %s", c.expectedTwap, twap), false
+	}
+
+	return ctx, nil, false
+}
+
+func CalcTwap(pair asset.Pair, twapCalcOpt v2types.TwapCalcOption, dir v2types.Direction, assetAmt sdk.Dec, twapLookbackWindow time.Duration, expectedTwap sdk.Dec) action.Action {
```
jgimeno, line 84: In this case I would call it TwapShouldBe, to make it clear that is some kind of assertion. It looks in this case more like an action that performs something.

k-yang: Done. Moved it to the `assertion` package also.

PR #1320, opened 2023-05-08
fix: localnet script

k-yang:
# Description

Adds peg-multiplier when creating perp vpools.

# Purpose

Fixes local testing


Unique-Divine:
Added a fix for the CosmWasm E2E contract tests

Unique-Divine (approved):

PR #1323, opened 2023-05-08
ci(deps): automatically update changelog with dependabot

Unique-Divine:
- Implements a CI workflow that allows dependabot to commit and update the changelog on its own whenever dependabot is labeled on a pull request.
- Adds the `github-actions` package ecosystem to dependabot.yml, which will check if our GH action versions are up to date


jgimeno:
Beautiful!

PR #1397, opened 2023-05-27
fix: ensure margin is high enough when removing it

matthiasmatt:
@k-yang how did you replace the openpositionshouldfail test? Unit tests are failing because one remove margin call i'm making is expected to fail and is effectively failing, but idk how to change that to make it expected


Unique-Divine (commented):
There's a lot of stuff commented out in the tests. Is this one still a draft?

Unique-Divine (approved):

PR #1407, opened 2023-06-01
feat!: upgrade to Cosmos SDK v47

jgimeno:
# Description

What does this PR do?

# Purpose

Why is this PR important?


contrib/devtools/Dockerfile
k-yang, line 1: Should we just replace the top-level Dockerfile?

contrib/devtools/Makefile
k-yang, line 1: Should we add this to the `contrib/make` directory?

contrib/scripts/protocgen2.sh
k-yang, line 1: Don't we already do this in `contrib/scripts/protogen.sh`?

jgimeno: done

k-yang (approved):

PR #1418, opened 2023-06-06
fix: fix build makefile

k-yang:
# Description

- Use the `uname -m` for determining computer arch instead of `uname -p`, which does not resolve to `x86_64` ever.

# Purpose

x86_64 machines were not able to build locally because it kept downloading the wrong librocksdb library.


AnishP15:
Works on my machine

Unique-Divine (approved):

PR #1421, opened 2023-06-12
feat(oracle): add expiry time to oracle prices

matthiasmatt:
# Description

What does this PR do?

# Purpose

Why is this PR important?


proto/oracle/v1/oracle.proto
```diff
@@ -130,6 +134,16 @@ message ExchangeRateTuple {
...
 }
 
+message DatedPrice {
+  string exchange_rate = 1 [
+    (gogoproto.moretags) = "yaml:\"exchange_rate\"",
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec",
+    (gogoproto.nullable) = false
+  ];
+
+  uint64 block_set = 2 [ (gogoproto.moretags) = "yaml:\"block_set\"" ];
```
k-yang, line 144: Maybe rename to `created_block`? `block_set` has the implication of a set data structure.

x/oracle/keeper/update_exchange_rates.go
```diff
@@ -84,11 +85,39 @@ func (k Keeper) getPairBallotsMapAndWhitelistedPairs(
...
+			err := k.ExchangeRates.Delete(ctx, key)
+			if err != nil {
+				panic(err)
+			}
+		}
+	}
+}
+
+// CleanExpiredRates removes all exchange rates from the state that are expired
+func (k Keeper) CleanExpiredRates(ctx sdk.Context) {
```
k-yang, line 108: Technically won't `resetExchangeRates` get rid of expired prices for us? So we don't need to call `CleanExpiredRates` at the end of every block, and we can re-use the iteration over the ballots at the end of every vote period instead (more efficient for computation).

k-yang: If we set expiration blocks as a multiple of voting period, then blocks will expire at the end of a voting period anyways so checking every block wouldn't be necessary. I think it would be worthwhile to save the compute, otherwise we could be slowing down our block times with a lot of state reads in EndBlocker.

k-yang (approved):

PR #1424, opened 2023-06-13
feat(perp): Add change type and exchanged margin to position changed events.

Unique-Divine:
# Purpose

Adds clarity on what funds were transferred to or from the trader. The fields added are especially useful for creating the perp leaderboard in the go-heartmonitor.


proto/perp/v2/event.proto
```diff
@@ -50,6 +50,20 @@ message PositionChangedEvent {
 
   // The block number at which this position was changed.
   int64 block_height = 7;
+
+  // Exchanged margin is the amount of collateral received by the trader during
+  // the position change. A positve value indicates that the trader received
+  // funds, while a negative value indicates that the trader spent funds.
+  string exchanged_margin = 8 [
```
k-yang, line 57: I would name this field `margin_to_user` to make clear what a positive or negative value means.

proto/perp/v2/event.proto
```diff
@@ -50,6 +50,20 @@ message PositionChangedEvent {
...
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int",
+    (gogoproto.nullable) = false
+  ];
+
+  // Change type describes the type of state transition that resulted in the
+  // position change. Change type can take the following values:
+  //
+  // ["open_position", "close_position", "liquidate", "add_margin", "remove_margin"]
+  string change_type = 9;
```
k-yang, line 66: I would use an enum here instead of a free-form string, to prevent data integrity errors downstream.

PR #1425, opened 2023-06-13
fix: remove positions from state when closed with reverse position

### LINKED ISSUE #1416: [AUDIT] Zero-Size Position Bug Distorts AMM Reserve/Short/Long Totals due to Erroneous DecreasePosition Flow Entry
* If you open a buy position then immediately open a sell position for the same amount (might be reachable in other ways), it's possible to enter the decreasePosition flow as the check at https://github.com/NibiruChain/nibiru/blob/d39099499d93c0d3ebdf3576f6cd937d6ea8d095/x/perp/v2/keeper/clearing_house.go#L223-L223 can pass (currentPositionNotional is fractionally larger) and you end up with a position with a size of 0. This position is then not saved due to the check at https://github.com/NibiruChain/nibiru/blob/d39099499d93c0d3ebdf3576f6cd937d6ea8d095/x/perp/v2/keeper/clearing_house.go#L564-L564, but the AMM is still updated as if it were saved (due to k.SwapQuoteAsset). This then messes up the reserve/short/long totals of the AMM:

matthiasmatt:
# Description

- fix #1416 

# Purpose

Why is this PR important?

### DIFF
--- a/x/perp/v2/keeper/clearing_house.go
+++ b/x/perp/v2/keeper/clearing_house.go
@@ -408,6 +408,12 @@ func (k Keeper) closeAndOpenReversePosition(
 	reverseNotionalValue := leverage.Mul(quoteAssetAmount)
 	remainingReverseNotionalValue := reverseNotionalValue.Sub(
 		closePositionResp.ExchangedNotionalValue)
+	if remainingReverseNotionalValue.IsNegative() {
+		// should never happen as openReversePosition should have checked this
+		return nil, nil, fmt.Errorf(
+			"provided quote asset amount and leverage not large enough to close position. need %s but got %s",
+			closePositionResp.ExchangedNotionalValue, reverseNotionalValue)
+	}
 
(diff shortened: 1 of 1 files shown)


Unique-Divine (commented):
Is their margin returned by the reverse close? And, is the position guaranteed to be empty if it has 0 size?

x/perp/v2/keeper/clearing_house.go
```diff
@@ -555,6 +555,11 @@ func (k Keeper) afterPositionUpdate(
 
 	if !positionResp.Position.Size_.IsZero() {
 		k.Positions.Insert(ctx, collections.Join(market.Pair, traderAddr), positionResp.Position)
+	} else {
+		if !positionResp.Position.Margin.IsZero() {
```
k-yang, line 559: Do we ever zero out the position's `margin`? I think we keep it there when we reverse a position.

k-yang (approved):

PR #1462, opened 2023-06-29
fix(perp): add pair to liquidation failed events

### LINKED ISSUE #1455: bug(perp): liquidation event is emitted twice for a single liquidation
Event itself is exactly the same.
[block_results_localnet.json.zip](https://github.com/NibiruChain/nibiru/files/11892520/block_results_localnet.json.zip)

### LINKED ISSUE #1456: bug(perp): MsgMultiLiquidateResponse is missing a pair
Can't determine a liquidated position without it.

Unique-Divine:
### Purpose 

- Closes #1456 
- Fixes #1455 

----

### Bonus

The sdk doesn't include many useful methods for testing events. This PR implements functions for: 
- ultra descriptive errors when asserts fail on the event manager. 
- easy parsing of typed events to JSON display
- a comparison function for checking fields of an `sdk.Event` without knowing its corresponding `proto.Message` (typed event)

The liquidation failed event was used as the Guinea pig:
 
wk27-n1700-TabTip_t4wb

feat: We may find it useful to base these methods on `proto.Message` types and generalize the concept for the `common/testutil` package.


x/perp/v2/keeper/liquidate.go
```diff
@@ -222,12 +240,12 @@ func (k Keeper) liquidate(
 		liquidationResponse.FeeToLiquidator,
 	)
 
-	ecosystemFundFee = sdk.NewCoin(
+	insuranceFundFee = sdk.NewCoin(
```
k-yang, line 243: Can we keep consistency in the names? It would be confusing if we use `ecosystemFund` in some places in the code and `insuranceFund` in other places.

x/perp/v2/keeper/liquidate.go
```diff
@@ -125,56 +132,65 @@ func PrettyLiquidateResponse(
...
+		eventLiqFailed := &types.LiquidationFailedEvent{
 			Pair:       pair,
 			Trader:     trader.String(),
 			Liquidator: liquidator.String(),
 			Reason:     types.LiquidationFailedEvent_NONEXISTENT_PAIR,
-		})
-		return sdk.Coin{}, sdk.Coin{}, types.ErrPairNotFound
+		}
+		_ = ctx.EventManager().EmitTypedEvent(eventLiqFailed)
+		err = fmt.Errorf("%w: pair: %s", types.ErrPairNotFound, pair)
```
k-yang, line 168: Let's use `Wrapf` instead of creating a new error, so we can preserve error codes in the response.

k-yang (changes requested):

x/perp/v2/keeper/liquidate.go
Unique-Divine: addressed by 4f854da4a8c575c5c2bc7263a14e6642c6778cfe

x/perp/v2/keeper/liquidate.go
Unique-Divine: - Moving this to separate PR  and adding the comment to ticket #1468

x/perp/v2/integration/assertion/event.go
```diff
@@ -14,18 +17,185 @@ import (
...
+func ProtoToJson(protoMsg proto.Message) (jsonOut string, err error) {
+	protoCodec := codec.NewProtoCodec(codectypes.NewInterfaceRegistry())
+	var jsonBz json.RawMessage
+	jsonBz, err = protoCodec.MarshalJSON(protoMsg)
+	return string(jsonBz), err
+}
+
+// EventEquals exports functions for comparing sdk.Events to concrete typed
+// events implemented as proto.Message instances in Nibiru.
+var EventEquals = eventEquals{}
```
k-yang, line 59: Does `EventEquals` need to be exported?

x/perp/v2/integration/assertion/event.go
```diff
@@ -14,18 +17,185 @@ import (
...
+type eventEquals struct{}
+
+// EventHasAttribueValue parses the given ABCI event at a key to see if it
+// matches (contains) the wanted value.
+//
+// Args:
+//   - abciEvent: The event under test
+//   - key: The key for which we'll check the value
+//   - want: The desired value
+func EventHasAttribueValue(abciEvent sdk.Event, key string, want string) error {
```
k-yang, line 70: Does `EventHasAttribueValue` need to be exported?

x/perp/v2/keeper/liquidate.go
```diff
@@ -18,44 +19,51 @@ import (
...
 
 	var allFailed bool = true
 
-	for i, req := range liquidationRequests {
-		cachedCtx, commit := ctx.CacheContext()
+	for reqIdx, req := range liquidationRequests {
+		// NOTE: writeCachedCtx (1) emits the events on the EventManager of the
+		// cachedCtx. and (2) writes that context to the commit multi store.
+
+		// cachedCtx, writeCachedCtx := ctx.CacheContext()
```
k-yang, line 30: We don't want to use cached ctx anymore?

k-yang (approved):

x/perp/v2/keeper/liquidate.go
Unique-Divine: Ah, good catch. I forgot to delete these comments. 

Removing the `ctx.CacheContext` stuff was the fix for the duplicate event issue that Oleg was encountering. 

I didn't really understand why we used the cached context before when I read this function. The event manager is a pointer that gets mutated throughout the execution, so it makes sense to append the failed liquidate messages onto it just like the successes.

x/perp/v2/integration/assertion/event.go
Unique-Divine: I'd honestly put it in testutil, but I wasn't sure where to put it

Issue #1412, opened 2023-06-04
[AUDIT] Zellic tracking issue [EPIC]


Unique-Divine:
## Main Task List

- [x] #1374
- [x] #1375
- [x] #1376
- [x] #1377 
- [x] #1378
- [x] https://github.com/NibiruChain/cw-nibiru/issues/31
- [x] #1415 
- [x] #1416 
- [x] #1438

---

## Updates

- 2023-06-14: Initial draft sent highlighting above issues
- 2023-06-15: Enter remediation section
- [x] Send commit hashes for fixes to Zellic

---

## Follow-up Changes 

Start Date: 2023-06-28

- [x] #1457
- [x] #1458

Unique-Divine:
Looks closed now

PR #1467, opened 2023-06-30
fix(oracle): make calcTwap safer

k-yang:
# Description

- removes potentially dangerous behavior of `calcTwap`, as pointed out by audit

# Purpose

- Audit: #1412


x/oracle/keeper/keeper.go
```diff
@@ -179,7 +130,26 @@ func (k Keeper) GetExchangeRateTwap(ctx sdk.Context, pair asset.Pair) (price sdk
...
+			// if we're at the last snapshot, then consider that price as ongoing until the current blocktime
+			nextTimestampMs = ctx.BlockTime().UnixMilli()
+		} else {
+			nextTimestampMs = snapshots[i+1].TimestampMs
+		}
+
+		price := s.Price.MulInt64(nextTimestampMs - s.TimestampMs)
+		cumulativePrice = cumulativePrice.Add(price)
+	}
+	return cumulativePrice.QuoInt64(ctx.BlockTime().UnixMilli() - firstTimestampMs), nil
```
Unique-Divine, line 152: Please add an explicit error that rules out zero division if the `snapshots[0].TimestampMs` is equal to the current block time

Unique-Divine (changes requested):

x/oracle/keeper/keeper.go
k-yang: Fixed.

Unique-Divine (approved):

PR #1492, opened 2023-07-05
test(perp): fix RemoveMargin test

k-yang:
# Description

`PositionShouldBeEqual` doesn't throw an error if the event is missing from the EventManager, so this PR addresses that. Then I realized that the `RemoveMargin` test had an extra assertion for checking that a `PositionChangedEvent` exists when it shouldn't.

# Purpose

Fixes perp event assertion testing.


k-yang:
Already merged in https://github.com/NibiruChain/nibiru/commit/27c75034a74e37adfc438c5cbc90588101d875c8

PR #1495, opened 2023-07-06
feat: add genmsg module

jgimeno:
# Description

What does this PR do?

# Purpose

Why is this PR important?


x/genmsg/genesis.go
```diff
@@ -0,0 +1,69 @@
...
+		handler := router.Handler(msg)
+		if handler == nil {
+			return fmt.Errorf("at index %d: no handler for message %T %s", i, msg, msg)
+		}
+		// resp, err := handler(context, msg)
+		_, err = handler(context, msg)
+		if err != nil {
+			return fmt.Errorf("at index %d: message processing: %w", i, err)
+		}
+		// log.Printf("message %d processed %s: %s", i, msg, resp.String())
```
Unique-Divine, line 66: I would take the response and pass and log it with the `ctx.Logger` here rather than ignoring the response. It will the genesis logs more informative.

x/genmsg/integration_test.go
```diff
@@ -0,0 +1,27 @@
+package genmsg_test
+
+import (
+	"testing"
+)
+
+func TestIntegration(t *testing.T) {
+	//recvAddr := sdk.AccAddress("recv")
```
Unique-Divine, line 8: You can use the helper in `common/testutil/cli` with the custom genesis, which will test your `InitGenesis` function. It's an easy way to create a happy path test for the whole module.

x/genmsg/module.go
```diff
@@ -0,0 +1,81 @@
...
+
+func (a AppModule) ExportGenesis(_ sdk.Context, codec codec.JSONCodec) json.RawMessage {
+	return codec.MustMarshalJSON(new(v1.GenesisState))
+}
+
+func (a AppModule) RegisterGRPCGatewayRoutes(_ client.Context, _ *runtime.ServeMux) {}
+
+func (a AppModule) GetTxCmd() *cobra.Command { return &cobra.Command{Use: ModuleName} }
+
+func (a AppModule) GetQueryCmd() *cobra.Command { return &cobra.Command{Use: ModuleName} }
```
Unique-Divine, line 71: Why are these blank commands?

contrib/make/post-genesis.mk
```diff
@@ -0,0 +1,2 @@
+post-genesis:
+	./contrib/scripts/post-genesis.sh
```
Unique-Divine, line 2: It looks like the script is missing from the pull request.

Unique-Divine (changes requested):

Unique-Divine (approved):

PR #1501, opened 2023-07-12
feat(scripts): python buf gen + localnet.sh improvements

Unique-Divine:
- feat(proto): python buf generation. This branch was used to create v0.21.x types for py-sdk
- docs: comments to help with running proto gen in python

- feat(localnet.sh): 
    1.  Makes script run while offline.
    2.  Adds `--no-build` flag option for external usage. 

Often when we use `localnet.sh` from outside this repo, there's no access to the rest of the source code to run `make install`, but we already have a versioned `nibid` binary. 

Binaries for `nibid` and `pricefeeder` can easily be installed with:
```bash
curl -s https://get.nibiru.fi/@v0.21.2! | bash
curl -s https://get.nibiru.fi/pricefeeder\! | bash
```

Adding the option to run:
```bash
bash scripts/localnet.sh --no-build
```
is very useful (1) when running e2e tests from outside this repo and (2) runs a lot faster since it skips the build and install step.


contrib/scripts/localnet.sh
```diff
@@ -34,14 +34,39 @@ echo_success() {
...
+do
+  if [ "$arg" == "--no-build" ] ; then
+    FLAG_NO_BUILD=true
+  fi
+done
+
+
+# Check if FLAG_NO_BUILD was set to true
+if ! $FLAG_NO_BUILD ; then
+  build_from_source
```
Unique-Divine, line 66: This is written in such a way that it should be easy to modify/extend with other flags in the future if the need arrises.

contrib/scripts/localnet.sh
```diff
@@ -139,15 +169,21 @@ else
...
+echo "$MNEMONIC" | $BINARY keys add $val_key_name --recover
+if $BINARY add-genesis-account $($BINARY keys show $val_key_name -a) $GENESIS_COINS; then
+  echo_success "Successfully added genesis account: $val_key_name"
 else
-  echo_error "Failed to add genesis accounts"
+  echo_error "Failed to add genesis account: $val_key_name"
 fi
 
+val_address=$($BINARY keys list | jq -r '.[] | select(.name == "validator") | .address')
+val_address=${val_address:-"nibi1zaavvzxez0elundtn32qnk9lkm8kmcsz44g7xl"}
```
Unique-Divine, line 183: Sets the `val_address` programmatically using `nibid keys list` rather than defaulting to a hardcoded one. That way, if we change the `$MNEMONIC` global variable, the script will still work.

contrib/scripts/localnet.sh
```diff
@@ -222,20 +261,33 @@ add_genesis_perp_markets_with_coingecko_prices() {
...
+add_genesis_param '.app_state.sudo.sudoers.root = "'"$val_address"'"'
 
-add_genesis_param '.app_state.oracle.params.twap_lookback_window = "900s"'
-add_genesis_param '.app_state.oracle.params.vote_period = "10"'
-add_genesis_param '.app_state.oracle.params.min_voters = "1"'
+# hack for localnet since we don't have a pricefeeder yet
+add_genesis_param '.app_state.oracle.exchange_rates[0].pair = "ubtc:unusd"'
+add_genesis_param '.app_state.oracle.exchange_rates[0].exchange_rate = "'"$price_btc"'"'
+add_genesis_param '.app_state.oracle.exchange_rates[1].pair = "ueth:unusd"'
+add_genesis_param '.app_state.oracle.exchange_rates[1].exchange_rate = "'"$price_eth"'"'
```
Unique-Divine, line 290: The hacky exchange rate values should match the initial mark prices set with `add_genesis_perp_markets_*`

PR #1527, opened 2023-07-27
test(common): docs and tests

Unique-Divine:
- refactor(testutil): organise network files
- refactor: DRY improvments for genesis cli tets
- test(common): document and test BankersRound
- fix(amm.go): small bug in ComputeSqrtDepth
- docs(testutil/cli): documentation for the Network and Validator structs
- test(common): address utils test


x/common/asset/pair.go
```diff
@@ -149,49 +149,19 @@ func (pairKeyEncoder) Decode(b []byte) (int, Pair) {
 	return i, MustNewPair(s)
 }
 
-//-----------------------------------------------------------------------------
```
Unique-Divine, line 152: Searched LSP and and found that this type alias was largely unused, so I removed some of the dead code.

x/common/dec.go
```diff
@@ -10,7 +10,7 @@ import (
```
Unique-Divine, line 10: Made the code more descriptive here and added tests for it. Since the `Sqrt` is used in `x/perp`, it's important to make sure these functions don't panic.

x/common/testutil/cli/logger.go
```diff
@@ -0,0 +1,16 @@
+package cli
```
Unique-Divine, line 1: Previously, the `Network`, `Config` (of the network), and `Logger` interface were all defined in one file, which made things hard to follow. I added documentation for parts that weren't obvious, separated these structs to their own files, added tests, and deleted code that was unused in integration tests.

x/common/testutil/cli/network_config.go
```diff
@@ -0,0 +1,63 @@
...
+	cfg.APIAddress = srvCfg.API.Address
+}
+
+func (cfg *Config) AbsorbTmConfig(tmCfg *tmconfig.Config) {
+	cfg.RPCAddress = tmCfg.RPC.ListenAddress
+}
+
+// AbsorbListenAddresses ensures that the listening addresses for an active
+// node are set on the network config.
+func (cfg *Config) AbsorbListenAddresses(val *Validator) {
```
Unique-Divine, line 60: The listen addresses for RPC, GRPC, and API should be set on the `Config` but weren't by default. This makes the API unintuitive when trying to grab the raw `grpc.Conn` objects or `rpc.RPCClient`.

PR #1529, opened 2023-07-28
feat: automatically tune GOMEMLIMIT and GOMAXPROCS

k-yang:
# Description

`github.com/KimMachineGun/automemlimit` is an imported package that automatically sets the `GOMEMLIMIT` runtime variable.

`go.uber.org/automaxprocs` is an imported package that automatically sets the `GOMAXPROCS` runtime variable.

Both of these variables tune the Golang runtime to prevent consuming too many resources. In particular, `GOMEMLIMIT` tells the Golang runtime to by default only use 90% of the detected environment's available memory.

# Purpose

Helps with memory leaks.


jgimeno:
In my opinion I would not add these kind of things to the node.

jgimeno (commented):
These are things that should be defined by the users who execute the binary. If I am not wrong you can modify it with some env variable before running the binary.

Unique-Divine (approved):

k-yang:
Closing since we can manually set these variables in our iac code.

PR #1541, opened 2023-08-11
feat: add clamp on funding rate

matthiasmatt:
# Description

What does this PR do?

# Purpose

Why is this PR important?

### DIFF
--- a/.github/workflows/proto-lint.yml
+++ b/.github/workflows/proto-lint.yml
@@ -22,7 +22,7 @@ jobs:
   #     timeout-minutes: 5
   #     steps:
   #       - uses: actions/checkout@v3
-  #       - uses: bufbuild/buf-setup-action@v1.25.1
+  #       - uses: bufbuild/buf-setup-action@v1.26.0
   #       - uses: bufbuild/buf-lint-action@v1
   #         with:
   #           input: "proto"
@@ -31,7 +31,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: bufbuild/buf-setup-action@v1.25.1
+      - uses: bufbuild/buf-setup-action@v1.26.0
       - uses: bufbuild/buf-breaking-action@v1
         with:
           input: "proto"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -96,7 +96,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Bump `github.com/spf13/viper` from 1.15.0 to 1.16.0 (#1436)
 - Bump `github.com/prometheus/client_golang` from 1.15.1 to 1.16.0 (#1431)
 - Bump `github.com/cosmos/ibc-go/v7` from 7.1.0 to 7.2.0 (#1445)
-- Bump `bufbuild/buf-setup-action` from 1.21.0 to 1.25.1 (#1449, #1469, #1505, #1510, [#1537](https://github.com/NibiruChain/nibiru/pull/1537))
+- Bump `bufbuild/buf-setup-action` from 1.21.0 to 1.26.0 (#1449, #1469, #1505, #1510, [#1537](https://github.com/NibiruChain/nibiru/pull/1537), [#1540](https://github.com/NibiruChain/nibiru/pull/1540))
 - Bump `google.golang.org/protobuf` from 1.30.0 to 1.31.0 (#1450)
 - Bump `cosmossdk.io/errors` from 1.0.0-beta.7 to 1.0.0 (#1499)
 - Bump `github.com/holiman/uint256` from 1.2.2 to 1.2.3 (#1504)


x/perp/v2/keeper/hooks.go
```diff
@@ -50,6 +50,12 @@ func (k Keeper) AfterEpochEnd(ctx sdk.Context, epochIdentifier string, _ uint64)
 		// See https://www.notion.so/nibiru/Funding-Payments-5032d0f8ed164096808354296d43e1fa for an explanation of these terms.
 		premiumFraction := markTwap.Sub(indexTwap).QuoInt64(int64(intervalsPerDay))
 
+		if premiumFraction.GT(market.MaxFundingRate) {
+			premiumFraction = market.MaxFundingRate
```
k-yang, line 54: I think we're mixing units here. Funding rate (in units of bps) can't be directly added to the premium fraction (units of $/contract). You have to multiply by index price to get premium fraction.

See https://www.notion.so/nibiru/Funding-Payments-5032d0f8ed164096808354296d43e1fa?pvs=4.

k-yang (changes requested):

proto/nibiru/perp/v2/state.proto
```diff
@@ -103,6 +103,15 @@ message Market {
   // the amount of collateral already credited from the ecosystem fund
   cosmos.base.v1beta1.Coin prepaid_bad_debt = 12
       [ (gogoproto.nullable) = false ];
+
+  // the maximum funding rate payment per epoch, this represents the maximum
+  // amount of funding that can be paid out per epoch as a percentage of the
+  // position size
+  string max_premium_fraction = 13 [
```
Unique-Divine, line 110: would prefer to say "max funding rate" since that's what we're enforcing here

x/perp/v2/keeper/hooks.go
```diff
@@ -50,6 +50,12 @@ func (k Keeper) AfterEpochEnd(ctx sdk.Context, epochIdentifier string, _ uint64)
 		// See https://www.notion.so/nibiru/Funding-Payments-5032d0f8ed164096808354296d43e1fa for an explanation of these terms.
 		premiumFraction := markTwap.Sub(indexTwap).QuoInt64(int64(intervalsPerDay))
 
+		if premiumFraction.GT(market.MaxPremiumFraction) {
+			premiumFraction = market.MaxPremiumFraction
+		} else if premiumFraction.LT(market.MaxPremiumFraction.Neg()) {
+			premiumFraction = market.MaxPremiumFraction.Neg()
+		}
```
Unique-Divine, line 57: Can you put this inside a function so it can be tested in isolation? I.e., 4 cases, pos + over cap, pos + under cap, neg + over cap, and neg + under cap

It would be easier for us to follow in the future than the app keeper tests

x/wasm/binding/cw_struct/cw_msg.go
```diff
@@ -133,6 +133,8 @@ type MarketParams struct {
 	PartialLiquidationRatio sdk.Dec `json:"partial_liquidation_ratio"`
 	// specifies the interval on which the funding rate is updated
 	FundingRateEpochId string `json:"funding_rate_epoch_id,omitempty"`
+	// specifies the maximum premium fraction to be paid out
+	MaxPremiumFraction sdk.Dec `json:"max_premium_fraction,omitempty"`
```
Unique-Divine, line 137: also "max_funding_rate"

Unique-Divine (changes requested):

x/perp/v2/keeper/hooks.go
k-yang: Yep, but the size needs to be converted to dollars too. For example, if I say the funding rate is 0.01%, then the funding payment is (0.01%)*(size)*(underlying price). We should rename to `funding_rate` and multiply by index price here.

k-yang: It's the first one.

Unique-Divine: You basically just need to clamp Divergence / num_payments_per_day

x/common/dec_test.go
```diff
@@ -129,3 +129,38 @@ func TestBankersRound(t *testing.T) {
...
+		value       sdk.Dec
+		clampValue  sdk.Dec
+		expected    sdk.Dec
+		description string
+	}{
+		{
+			value:       sdk.NewDec(15),
+			clampValue:  sdk.NewDec(1),
+			expected:    sdk.NewDec(1),
+			description: "Clamping 1.5 to 1",
```
k-yang, line 144:
```suggestion
			description: "Clamping 15 to 1",
```

x/common/dec_test.go
```diff
@@ -129,3 +129,38 @@ func TestBankersRound(t *testing.T) {
...
+		},
+		{
+			value:       sdk.NewDec(-15),
+			clampValue:  sdk.NewDec(1),
+			expected:    sdk.NewDec(-1),
+			description: "Clamping -1.5 to 1",
```
k-yang, line 150:
```suggestion
			description: "Clamping -15 to 1",
```

x/perp/v2/client/cli/gen_market.go
```diff
@@ -167,6 +167,7 @@ func newMarketFromFlags(flagSet *flag.FlagSet,
 		LiquidationFeeRatio:             sdk.MustNewDecFromStr("0.0500"),
 		PartialLiquidationRatio:         sdk.MustNewDecFromStr("0.5"),
 		FundingRateEpochId:              epochstypes.ThirtyMinuteEpochID,
+		MaxFundingRate:                  sdk.NewDec(1),
```
k-yang, line 170: Are you sure you want the default max funding rate to be 100%?

k-yang (approved):

Unique-Divine (approved):

PR #1565, opened 2023-09-05
fix(oracle)!: count vote omission as abstain for less slashing + more stability

Unique-Divine:
- changelog
- fix(oracle): miss outside reward band but not on abstain
- fix(oracle): #wip checkpoint for omit == abstain

# Description

What does this PR do?

# Purpose

Why is this PR important?


x/oracle/keeper/test_utils.go
```diff
@@ -285,8 +285,8 @@ func AllocateRewards(t *testing.T, input TestFixture, rewards sdk.Coins, votePer
 }
 
 var (
-	stakingAmt         = sdk.TokensFromConsensusPower(10, sdk.DefaultPowerReduction)
-	randomExchangeRate = sdk.NewDec(1700)
+	TEST_STAKING_AMT = sdk.TokensFromConsensusPower(10, sdk.DefaultPowerReduction)
```
k-yang, line 288: Let's avoid global variables in tests.

x/oracle/keeper/slash_test.go
```diff
@@ -160,46 +160,72 @@ func TestInvalidVotesSlashing(t *testing.T) {
...
+		perfs := input.OracleKeeper.UpdateExchangeRates(input.Ctx)
+		missCount := input.OracleKeeper.MissCounters.GetOr(input.Ctx, ValAddrs[0], 0)
+		require.EqualValues(t, 0, missCount, perfs.String())
+	}
+
+	t.Log("valoper0 should not be slashed")
+	validator := input.StakingKeeper.Validator(input.Ctx, ValAddrs[0])
+	require.Equal(t, TEST_STAKING_AMT, validator.GetBondedTokens())
+
+	// // one more miss vote will inccur Account 1 slashing
```
k-yang, line 216: Should these tests be commented out?

k-yang:
Overall I don't think we should make this change because it changes the economic incentives for validators to participate in the decentralized oracle. If the concern is about network stability, I think we should analyze what went wrong with the smaller slashing window and higher min_valid_pct instead of immediately making this change.

x/oracle/keeper/test_utils.go
Unique-Divine: It's just a rename to make it more obvious that we're referencing global vars. Should we just add them as test variables to the struct?

x/oracle/keeper/slash_test.go
Unique-Divine: Yes, meant to just delete this part since you don't get slashed for abstaining

x/oracle/keeper/test_utils.go
Unique-Divine: Or is it sufficient to just lowercase the `T` to make it private?

x/oracle/keeper/slash_test.go
```diff
@@ -160,46 +160,47 @@ func TestInvalidVotesSlashing(t *testing.T) {
...
+			valIdx)
+	}
+	input.OracleKeeper.WhitelistedPairs.Insert(input.Ctx, pair)
+	perfs := input.OracleKeeper.UpdateExchangeRates(input.Ctx)
+	require.EqualValues(t, 0, perfs.GetTotalRewardWeight())
+
+	allowedMissPct := sdk.OneDec().Sub(minValidPerWindow)
+	allowedMissPeriods := allowedMissPct.MulInt64(votePeriodsPerWindow).
+		TruncateInt64()
+	t.Logf("For %v blocks, valoper0 does not vote, while 1 and 2 do.", allowedMissPeriods)
```
k-yang: I think you're conflating blocks with vote periods here. One vote period is many blocks (by default 30 on testnet), so if the goal is to vote once every vote period, we have to skip the block number by 30 in each iteration.

However, this test doesn't really test validator slashing, only that the MissCounter doesn't increment. The slashing happens in the abci `EndBlocker` which is in another test.

x/oracle/keeper/test_utils.go
k-yang: It's more about not creating globally mutable variables. It's a bad practice because it creates the possibility of accidental mutation and which could have unknown downstream consequences.

x/oracle/types/ballot.go
```diff
@@ -173,9 +174,14 @@ func (pb ExchangeRateBallots) Swap(i, j int) {
...
-	Power        int64 // tendermint consensus power
-	RewardWeight int64 // how much of the rewards this validator should receive, units of consensus power
-	WinCount     int64
+	// Tendermint consensus voting power
+	Power int64
+	// RewardWeight: Weight of rewards the validator should receive in units of
+	// consensus power.
+	RewardWeight int64
+	WinCount     int64 // Number of valid votes for which the validator will be rewarded
+	AbstainCount int64 // Number of abstained votes for which there will be no reward or punishment
```
k-yang, line 183: I don't think we're using AbstainCount usefully anywhere. Let's just get rid of it?

x/oracle/keeper/update_exchange_rates.go
```diff
@@ -27,21 +27,43 @@ func (k Keeper) UpdateExchangeRates(ctx sdk.Context) {
...
+				ctx, validatorPerformance.ValAddress,
+				k.MissCounters.GetOr(ctx, validatorPerformance.ValAddress, 0)+1)
+			k.Logger(ctx).Info("vote miss", "validator", validatorPerformance.ValAddress.String())
 		}
+	}
+}
 
-		k.MissCounters.Insert(ctx, validatorPerformance.ValAddress, k.MissCounters.GetOr(ctx, validatorPerformance.ValAddress, 0)+1)
-		k.Logger(ctx).Info("vote miss", "validator", validatorPerformance.ValAddress.String())
+func (k Keeper) registerAbstainsByOmission(
```
k-yang, line 51: I don't think we need this function. The return type of `UpdateExchangeRates` isn't used anywhere so we don't need to do anything after rewarding validators and updating miss counters.

x/oracle/keeper/slash_test.go
Unique-Divine: Yes, the name is wrong because it used to be a test about slashing that had abstain votes. Since that doesn't cause slashing, the test is no longer about that either. Will need to repurpose it

x/oracle/keeper/test_utils.go
Unique-Divine: Test variables are only available in the package_test namespace, so they can't be used in the normal business logic

x/oracle/types/ballot.go
Unique-Divine: There's a ticket open to as an event for this

x/oracle/keeper/ballot.go
```diff
@@ -137,11 +137,17 @@ func (k Keeper) removeInvalidBallots(
...
 //
-// ALERT: This function mutates validatorPerformances slice based on the votes made by the validators.
-func Tally(ballots types.ExchangeRateBallots, rewardBand sdk.Dec, validatorPerformances types.ValidatorPerformances) sdk.Dec {
+// ALERT: This function mutates validatorPerformances slice based on the votes
+// made by the validators.
+func Tally(
+	ballots types.ExchangeRateBallots,
+	rewardBand sdk.Dec,
+	validatorPerformances types.ValidatorPerformances,
+) (sdk.Dec, types.ValidatorPerformances) {
```
jgimeno, line 150: This is a bad idea, to have a return only used for tests. The function already mentions that it mutates the validator perfomances, returning it is confusing.

k-yang (approved):

PR #1575, opened 2023-09-13
feat(perp): Add TraderVolume tracking

testinginprod:
# Description

Adds a way to track the user volume on successful position changes.

- Closes: https://github.com/NibiruChain/nibiru/issues/1576

# Purpose

Enables discounts and rebates.

- Related: #1587


proto/nibiru/perp/v2/genesis.proto
```diff
@@ -19,4 +19,15 @@ message GenesisState {
...
+  repeated TraderVolume trader_volumes = 7 [ (gogoproto.nullable) = false ];
+
+  message TraderVolume {
+    string trader = 1;
+    uint64 epoch = 2;
+    string volume = 3 [
+      (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int",
+      (gogoproto.nullable) = false
+    ];
+  }
```
Unique-Divine, line 32: I'd probably leave `TraderVolume` out of the genesis so that it starts out empty and just gets populated from trades. I don't see a big use case for starting someone's volume out nonzero.

x/perp/v2/keeper/dnr.go
```diff
@@ -0,0 +1,90 @@
...
+	return value.String()
+}
+
+func (i intValueEncoder) Name() string {
+	return "math.Int"
+}
+
+// IncreaseTraderVolume adds the volume to the user's volume for the current epoch.
+func (k Keeper) IncreaseTraderVolume(ctx sdk.Context, user sdk.AccAddress, volume math.Int) {
+	currentEpoch, err := k.DnREpoch.Get(ctx)
```
Unique-Divine, line 45: refactor suggestion: I would pass in `currentEpoch` as an argument and move the getter to be called just before this function for two reasons:
1. You'll need the `currentEpoch` value to compute the taker fee on this part of clearing_house.go:
```go
	// update user volume
	k.IncreaseTraderVolume(ctx, traderAddr, positionResp.ExchangedNotionalValue.Abs().TruncateInt())

	transferredFee, err := k.transferFee(ctx, market.Pair, traderAddr, positionResp.ExchangedNotionalValue,
		market.ExchangeFeeRatio, market.EcosystemFundFeeRatio,
	)
```
Here (in a future PR), you'll need the `currentEpoch` value to get the trader's DnR table

2. When you run this in a loop for garbage collection, you can just do 1 state read for the epoch and re-use the value while iterating through the trader addresses to make it a bit more efficient

Unique-Divine (changes requested):

proto/nibiru/perp/v2/genesis.proto
Unique-Divine: Ah, I see

Unique-Divine (approved):

PR #1606, opened 2023-09-27
fix(perp): emit MarketUpdatedEvents even in the absence of index price

### LINKED ISSUE #1561: fix: `MarketUpdated` and `AmmUpdated` events should still be published in the absence of an index price

k-yang:
# Description

Emit `MarketUpdatedEvent` and `AmmUpdatedEvent` even if there is no valid index price.

# Purpose

- Closes #1561


x/perp/v2/module/abci.go
```diff
@@ -37,18 +37,19 @@ func EndBlocker(ctx sdk.Context, k keeper.Keeper) []abci.ValidatorUpdate {
...
 			k.Logger(ctx).Error("mark price is zero", "market.Pair", market.Pair)
 			continue
 		}
 
-		if indexTwap.IsNil() || indexTwap.IsZero() {
+		var indexTwap sdk.Dec
+		indexTwap, err = k.OracleKeeper.GetExchangeRateTwap(ctx, amm.Pair)
+		if err != nil {
+			k.Logger(ctx).Error("failed to fetch twap index price", "market.Pair", market.Pair, "error", err)
+			indexTwap = sdk.ZeroDec()
```
Unique-Divine, line 49: Since we consider zero price as an error case, could we maybe use negative one as the error value instead to make it super obvious that it's invalid for clients reading the events?

k-yang: Done.

Unique-Divine (approved):

Issue #1584, opened 2023-09-14
[epic] feat: "tokenfactory" module


Unique-Divine:
## Context

Implement a permissionless `tokenfactory` module based on the standard used by Osmosis, Juno, Injective, and other networks.

> "This is important for cosmwasm applications on osmosis to thrive." - [osmosis#1029](https://github.com/osmosis-labs/osmosis/issues/1029)

## Rationale

- Sending CW20 asssets that use the factory can be simple
  "/cosmos.bank.v1beta1.MsgSend" tx msgs. This makes it much easier to transfer tokens at the CLI without invoking multiple wasm contracts. 
- The `x/bank` module is well-tested and exists on every Cosmos-SDK chain, whereas CW20 only exists on chains with CosmWasm.
- All smart contracts can access the bank module messages and queries by default, and this will continue to be the case even if we have ERC-20 assets with EVM support next year.
- Way more efficient query via x/bank with built-in total supply tracking. No need to do paginated CW20 queries and then filter the pages.
- Factory tokens can automatically be displayed in popular tools like explorers (Mintscan), Leap, Keplr, etc.
- Saves us from needing to migrate to the `bank` standard in the future like [Juno](https://github.com/CosmosContracts/tokenfactory-contracts/tree/main/contracts/migrate).

## Completion Tracking

- [x] Impl `tokenfactory/types`
- [x] Impl the `tokenfactory` state with collections
- [x] Impl `tokenfactory/keeper` business logic (and wire module to app)
- [x] Impl CLI for `tokenfactory`
- [x] Impl [wasm bindings for `tokenfactory`](https://github.com/osmosis-labs/osmosis/pull/1135#issue-1178600217)
- [x] test(cw-nibiru): Create an e2e test using the `tokenfactory` wasm bindings

## Core Functionality

For smart contracts, the Public API for the transaction messages of this module could look something like this. 

```rust
use cosmwasm_std::Int256;
type SdkInt = Int256; // alias for cosmos-sdk/types.Int

/// `TokenfactoryExecuteMsg` describes the different messages that the token
/// factory can accept.
enum TokenfactoryExecutMsg {
    /// CreateDenom creates a token factory denomination of the form: 
    /// "factory/{contract-address}/{subdenom}".
    /// Contracts can create denoms, namespaced under the contract's address.
    /// A contract may create any number of independent sub-denoms.
    CreateDenom {
        /// `subdenom` can be of length at most 44 characters, in [0-9a-zA-Z./]
[...]

PR #1607, opened 2023-09-28
feat(tokenfactory): execute msgs for CreateDenom, ChangeAdmin, UpdateModuleParams

Unique-Divine:
- Includes queries, transaction messages, and CLI commands for CreateDenom, ChangeAdmin, and UpdateModuleParams
- Part of #1584 
- Includes fix from #1612 (preferably, merge this one first to avoid conflicts)


Unique-Divine:
I'm encountering something strange here where it says the perp protos are breaking and I'm not even editing those files.

image

image

x/tokenfactory/keeper/grpc_query.go
```diff
@@ -38,3 +39,47 @@ func (q Querier) Params(
 	}
 	return &types.QueryParamsResponse{Params: params}, nil
 }
+
+// Denoms: Returns all registered denoms for a given creator.
+func (q Querier) Denoms(
```
jgimeno, line 44: I dont see why we need this and the other one, they are very similar

x/tokenfactory/keeper/msg_server.go
```diff
@@ -0,0 +1,97 @@
...
+	grpcstatus "google.golang.org/grpc/status"
+
+	"github.com/NibiruChain/nibiru/x/tokenfactory/types"
+)
+
+var _ types.MsgServer = (*Keeper)(nil)
+
+var errNilMsg error = grpcstatus.Errorf(grpccodes.InvalidArgument, "nil msg")
+
+func (k Keeper) CreateDenom(
```
jgimeno, line 18: I think we created this without thinking the use case and how it aligns with our chain.

x/tokenfactory/keeper/grpc_query.go
Unique-Divine: It's the DenomsFromCreator query

jgimeno (approved):

PR #1614, opened 2023-10-01
refactor(proto): Use explicit namespacing on proto imports for #1608

### LINKED ISSUE #1608: refactor(x): Use explicit namespacing on proto imports
## Context

Most of out proto imports look something like this: 

We have a proto file such as `devgas/v1/tx.proto` that imports from `devgas/v1/genesis.proto`.

```proto
package nibiru.devgas.v1;
import "nibiru/devgas/v1/genesis.proto";

message MsgUpdateParams {
  // ... other fields

  ModuleParams params = 2 [ (gogoproto.nullable) = false ];
}
```

Here, `ModuleParams` is defined in `genesis.proto`. however, it's not obvious where the type comes from when reading the protos. Futhermore, this type can have a name collision with another type like `perp/.../ModuleParams`.

Protobuf allows for explicitly namespacing types that are imported from other files. Thus, we can write:

```proto
package nibiru.devgas.v1;
import "nibiru/devgas/v1/genesis.proto";

message MsgUpdateParams {
  // ... other fields

  nibiru.devgas.v1.ModuleParams params = 2 [ (gogoproto.nullable) = false ];
}
```

I think explicitly writing out the import location will improves readbility a lot here, especially when a proto file imports types from multiple places.

## Proposal

Do this in every module:

- [x] perp
- [x] inflation
- [x] spot
- [x] oracle
- [x] epochs
- [x] devgas
- [x] tokenfactory
- [x] sudo
- [x] genmsg

Unique-Divine:
- Closes #1608
- Deletes `proto/nibiru/stablecoin` since we're no longer using those types.
- This change is completely non-breaking. It doesn't affect the Golang, just the the imports in the protobuf files.


jgimeno:
Why we do this random stuff? It is not useful and adds noise

jgimeno (commented):
no, I dont see why this needs to be done

Issue #1588, opened 2023-09-18
feat(ante): set minimum commission percentage for validating


Unique-Divine:
Since we're using Cosmos-SDK v0.47+, we can just set this in the `x/staking` module params.

- See https://github.com/cosmos/cosmos-sdk/pull/10529#issuecomment-1026320612

PR #1615, opened 2023-10-01
feat(ante)!: Ante handler to add a maximum commission rate of 25% for validators.

Unique-Divine:
# Purpose

In our validator interviews, we found that a common strategy for "low-commission attacking" validators is to:

1. Set a normal commission rate
2. Then, change the commission rate to be extremely low to undercut the other validators and advertise for delegations
3. And finally, adjust the commission extremely high to rake in profits with the new delegations.

If you impose a reasonable minimum and maximum commission, this problem is almost entirely avoided.

- Related to #1588

### DIFF
--- a/app/ante/fixed_gas_test.go
+++ b/app/ante/fixed_gas_test.go
@@ -3,9 +3,9 @@ package ante_test
 import (
 	"testing"
 
-	"github.com/cosmos/cosmos-sdk/types/errors"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 
-	sdkerrors "cosmossdk.io/errors"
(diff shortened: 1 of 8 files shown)


app/ante/fixed_gas_test.go
```diff
@@ -196,7 +196,7 @@ func (suite *AnteTestSuite) TestOraclePostPriceTransactionsHaveFixedPrice() {
 					Amount:      sdk.NewCoins(sdk.NewInt64Coin(app.BondDenom, 200)),
 				},
 			},
-			expectedGas: 62504,
+			expectedGas: 62288,
```
Unique-Divine, line 199: Somehow, gas decreases due to this change. I have no idea why

PR #1620, opened 2023-10-03
feat(tokenfactory): tx msgs for mint and burn

Unique-Divine:
- Includes transaction messages and CLI commands for `MsgMint` and `MsgBurn`
- Implements and tests `MsgSetDenomMetadata`
- Part of #1584 
- Next step: Wasm bindings


k-yang (approved):
Approved. Would it be possible to split big PRs like this into separate smaller PRs in the future? That would make them easier to review.

An example here could be a PR for `mint`, `burn`, and `denom metadata`.

PR #1630, opened 2023-10-09
refactor(wasm): wasmbinding folder structure

k-yang:
# Description

- re-organizes the wasmbinding/ folder to be more alike the idiomatic structure followed by [Neutron](https://github.com/neutron-org/neutron/tree/main/wasmbinding), [Osmosis](https://github.com/osmosis-labs/osmosis/tree/main/wasmbinding), and [Archway](https://github.com/archway-network/archway/tree/main/wasmbinding)
- renames some type names for clarity

# Purpose

Having our codebase be similar in structure to canonical examples in the Cosmos ecosystem makes it easier for external developers to understand our codebase and for internal developers to understand external codebases without much context switching.


jgimeno (approved):

Unique-Divine:
Putting the files in the top level instead of `x/wasm` seems to break the module convention

k-yang:
> Putting the files in the top level instead of `x/wasm` seems to break the module convention

wasmbindings are not an application module though. I would consider it as part of the chain infra, which could deserve it's own top-level directory.

Looking at examples from Osmosis, Archway, and Neutron, they all seem to follow this convention. The actual wasm module comes from wasmd/x/wasm, which is imported in go.mod.

k-yang:
> Yes, but our import of wasm includes stuff from our `x/wasm` folder. Just saying, other chains doing something doesn't automatically mean it's a convention we should follow

I updated the wasmbinding imports in app.go, and verified all tests pass.

My reasoning for following the same convention as other chains is that it reduces cognitive complexity if different codebases follow the same structure. If you want to keep the module convention though, I don't mind putting it under x/wasmbinding.

Unique-Divine:
> > Yes, but our import of wasm includes stuff from our `x/wasm` folder. Just saying, other chains doing something doesn't automatically mean it's a convention we should follow
> 
> I updated the wasmbinding imports in app.go, and verified all tests pass.
> 
> My reasoning for following the same convention as other chains is that it reduces cognitive complexity if different codebases follow the same structure. If you want to keep the module convention though, I don't mind putting it under x/wasmbinding.

Would prefer this or even `x/wasm/wasmbinding`

PR #1632, opened 2023-10-10
feat: create settle position function

matthiasmatt:
# Description

- [x] waiting to review codecov report

# Purpose

Why is this PR important?


Unique-Divine (approved):
LGTM, I am a bit curious about how we can handle the fact that there's a variable part that could be unexpectedly large for the funding payments. 

Also, when we close one market, how are the conditions for the _next_ market selected?

PR #1639, opened 2023-10-17
fix(perp): by default, new markets are disabled

k-yang:
# Description

What does this PR do?

# Purpose

Why is this PR important?

### DIFF
--- a/wasmbinding/exec_test.go
+++ b/wasmbinding/exec_test.go
@@ -550,7 +550,7 @@ func (s *TestSuiteExecutor) TestCreateMarket() {
 	market, err := s.nibiru.PerpKeeperV2.GetMarket(s.ctx, pair)
 	s.NoError(err)
 	s.NoError(market.Validate())
-	s.True(market.Enabled)
+	s.False(market.Enabled, "by default, we create a market in a disabled state")
 	s.EqualValues(pair, market.Pair)
 
 	s.T().Log("Executing without permission should fail")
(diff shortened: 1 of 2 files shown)


x/common/testutil/action/testcase.go
```diff
@@ -48,67 +48,73 @@ func TC(name string) TestCase {
...
-
-func (t *TestSuite) WithTestCases(testCase ...TestCase) *TestSuite {
-	t.testCases = append(t.testCases, testCase...)
-	return t
-}
-
-func (t *TestSuite) Run() {
-	for _, testCase := range t.testCases {
+func (tc TestCase) Run(t *testing.T) {
+	t.Run(tc.Name, func(t *testing.T) {
```
Unique-Divine, line 67: All that's happening here is that the execution block is wrapped inside of a `t.Run`. This makes the test logs less opaque because it clarifies further which specific test case is failing. 

One issue with out stack trace right now is that it always shows a failing test as coming from `testcase.go`.

x/perp/v2/keeper/settlement.go
```diff
@@ -7,9 +7,10 @@ import (
 	"github.com/NibiruChain/nibiru/x/perp/v2/types"
 )
 
-// CloseMarket closes the market. From now on, no new position can be opened on this market or closed.
-// Only the open positions can be settled by calling SettlePosition.
-func (k Keeper) CloseMarket(ctx sdk.Context, pair asset.Pair) (err error) {
+// CloseMarket closes the market. From now on, no new position can be opened on
+// this market or closed. Only the open positions can be settled by calling
+// SettlePosition.
+func (k admin) CloseMarket(ctx sdk.Context, pair asset.Pair) (err error) {
```
k-yang, line 13: Should we move this to the `admin.go` file?

Unique-Divine: Yeah, that would be more consistent with out pattern

Issue #1580, opened 2023-09-14
feat(wasm): Create a Rust package to call module TxMsgs with `CosmosMsg::StargateMsg`


Unique-Divine:
Tracking issue for migrating wasm bindings to use `CosmosMsg::StargateMsg`

## Higher Prio
- [ ] Generate Rust proto types
- [ ] Create Stargate messages for `x/tokenfactory`  (with tests)
- [ ] Create Stargate queries for `x/oracle` (with tests)

## Nice to Have
- [ ] Create Stargate messages for `x/perp` 
- [ ] Create Stargate messages for `x/spot`

Unique-Divine:
Update: The framework for completing this ticket is implemented inside the `nibiru_std::proto` crate with the `NibiruStargateMsg` and `NibiruStargateQuery` traits. Several trait implementations have already been added in files like [proto/type_url_nibiru.rs](https://github.com/NibiruChain/cw-nibiru/blob/becc614b5a1b51f3da2e1646a3f26be91340627c/nibiru-std/src/proto/type_url_nibiru.rs).

To complete the remainder of this task, `impl` blocks for `Prost::Name` need to be added for the above checklist, which is relatively straightforward.

PR #1646, opened 2023-10-20
feat(wasmbinding)!: whitelisted stargate queries for QueryRequest::Stargate: auth, bank, gov, tokenfactory, epochs, inflation, oracle, sudo, devgas

Unique-Divine:
# Purpose

- Adds whitelisted stargate queries for `QueryRequest::Stargate` for the following modules: bank, auth, gov, tokenfactory, epochs, inflation, oracle, sudo, devgas.
- Fixes inconsistent protobuf names on `EpochInfos`
- Partially completes #1580

Why is this PR important?

Without this, it's not possible to use `QuerySupplyOf` from `x/bank` or fetch denom metadata from a smart contract.


x/stablecoin/types/events.pb.go
Unique-Divine, line 1: How do we keep merging this? lol

Unique-Divine:
Note that the buf break is the typo fix

wasmbinding/stargate_query_test.go
```diff
@@ -0,0 +1,143 @@
...
+request or response, we should know the `QueryRequest::Stargate.path`
+deterministically.
+*/
+func TestWasmAcceptedStargateQueries(t *testing.T) {
+	stargateQueryPaths := set.New[string]()
+	t.Log("stargateQueryPaths: Add cosmos and ibc query paths")
+	// These are hard-coded because the GRPC service descriptions aren't exported
+	// from the Cosmos-SDK and remain private vars. Maybe we could ask the
+	// maintainers to export them in the future.
+	for _, queryPath := range []string{
```
k-yang, line 45: Maybe it's better to iterate over the return map of `WasmAcceptedStargateQueries()` instead of having a duplicate list of hard-coded strings.

wasmbinding/stargate_query_test.go
```diff
@@ -0,0 +1,143 @@
...
+		isExceptionPath := exceptionPaths.Has(pbQueryResponseTypeUrl)
+		splitResponse := strings.Split(pbQueryResponseTypeUrl, "Response")
+		assert.Lenf(t, splitResponse, 2, "typeUrl: %v",
+			splitResponse, pbQueryResponseTypeUrl)
+
+		// Get proto message "package" from the response type
+		typeUrlMinusSuffix := splitResponse[0]
+		typeUrlPartsFromProtoMsg := strings.Split(typeUrlMinusSuffix, ".")
+		assert.GreaterOrEqual(t, len(typeUrlPartsFromProtoMsg), 4, typeUrlPartsFromProtoMsg)
+		protoMessagePackage := typeUrlPartsFromProtoMsg[:3]
```
k-yang, line 117: I'm not sure if 3 is the right amount in all cases. For example, the IBC core ones are `ibc.core.connection.v1`.

wasmbinding/stargate_query_test.go
```diff
@@ -0,0 +1,143 @@
...
+		// Verify that the packages match
+		assert.Equalf(t, queryPathProtoPackage, protoMessagePackage,
+			"package names inconsistent:\nfrom query path: %v\nfrom protobuf object: %v",
+			queryPath, pbQueryResponseTypeUrl,
+		)
+
+		// Verify that the method names match too.
+		methodNameFromPb := strings.TrimLeft(typeUrlPartsFromProtoMsg[3], "Query")
+		methodNameFromPath := strings.TrimLeft(typeUrlPartsFromQueryPath[3], "Query/")
+		if !isExceptionPath {
```
k-yang, line 133: This could be moved up to line 109
```go
if isExceptionPath {
    continue
}
```

Prefer the "[early return pattern](https://medium.com/swlh/return-early-pattern-3d18a41bba8)" over logic wrapped in branches.

x/inflation/types/export.go
```diff
@@ -0,0 +1,12 @@
+package types
+
+import (
+	grpc "google.golang.org/grpc"
+)
+
+// GrpcQueryServiceDesc represents the query server's RPC service specification.
+// This gives access to the service name and method names needed for stargate
+// queries.
+func GrpcQueryServiceDesc() grpc.ServiceDesc {
```
k-yang, line 10: I'm a little bit worried about exposing the service descriptor as a public variable because it allows for mutations from other places in our codebase, which could mess up the gRPC query server. I believe the query descriptors were made package-private for a reason.

Would it be possible to move the typeURL tests into the packages themselves and not expose the query descriptors? You'll still have access to the list of fully qualified service method names and protobuf names from the `wasmbinding.WasmAcceptedStargateQueries()` function.

Unique-Divine: I'll edit the exported function to return a copy, which will make it more defensive

Unique-Divine: Actually, now that I'm looking at it more closely, the `grpc.ServiceDesc` isn't using pointers, so the underlying value used in the module is immutable, which is what we want.

```go
func GrpcQueryServiceDesc() grpc.ServiceDesc {
	return _Query_serviceDesc
}
```

Since we're exporting as a function, I don't think there's any way to modify the actual `_Query_serviceDesc`

wasmbinding/stargate_query_test.go
Unique-Divine: Yes, good idea. Addressed in 6f4b57767b63a222982e1e2bfa27b215e8d7c570

wasmbinding/stargate_query_test.go
Unique-Divine: I think you're right. For on where we assert >= instead of >, the package would be all except that last element rather than the first 3. 

Addressed in a69b501ed12e17c48e58e254e62efae239956588

wasmbinding/stargate_query_test.go
Unique-Divine: Good idea. It makes it more concise too. Updated in 6f4b57767b63a222982e1e2bfa27b215e8d7c570

k-yang (approved):

PR #1651, opened 2023-10-26
feat(openapi): generate openapi.yml with Ignite CLI for use with Swagger

### LINKED ISSUE #761: docs: generate vuejs openapi docs for the chain
Using [ignite CLI](https://docs.ignite.com/guide/install.html#upgrading-your-ignite-cli-installation), you can run this to get the needed  YAML:

```sh
ignite generate openapi
```

From there we need to use a tool (Swagger or something similar) to generate the server.

Unique-Divine:
## Purpose 

- Closes #761  (mostly). Technically, this just creates the Open API spec, but that's the tricky part of the ticket

## For Review

Just run 
```bash
bash tools/ignite-openapi.sh
```

and you should get an output like so:

```bash
✅ Success: symlinked config
Checking chain config file:
✔ config file OK
Checking dependency tools:
✔ tools/tools.go exists
✔ tools file OK
✔ Generated OpenAPI spec
⋆ OpenAPI path: /home/realu/warpath/NibiruChain/nibiru/dist/openapi.yml
✅ Success: Completed generation of OpenAPI spec
```

Note that the output file is more than 27,000 lines long, so we shouldn't commit it.


config.yml
Unique-Divine, line 1: This is a symlink. Git has a special way of representing these.

tools/tools.go
Unique-Divine, line 1: This is auto-generated by ignite CLI

Issue #1623, opened 2023-10-04
test: How can we make the CI integration test runs more deterministic?


Unique-Divine:
## Context

Test runs on the same code are very finicky because our `cli.Network` integration tests create several files in temporary test directories and use real JSON-RPC and GRPC connections for queries and transactions. 

For example, many of the following GitHub action runs (2023-10-04) are on passing code but fail with network errors in unchanged modules:

![image](https://github.com/NibiruChain/nibiru/assets/51418232/4181fb50-163e-41ae-8591-9e1253f102ab)

## Some Ideas on Potential Solutions

1. More clever utilization of mutex?
2. Calling `Network.WaitForNextBlock()` after each `ExecTx`
3. As a rule, spin up at most 1 network per suite to make them more lightweight

## Tasks 

- [ ] Prevent [Common Error 1: tcp binding address in use][common-error-1]
- [ ] Prevent [Common Error 2: spot tx hash not found][common-error-2]

[common-error-1]: https://github.com/NibiruChain/nibiru/issues/1623#common-foo-bar:~:text=hash%20not%20found-,Common%20Error%201,-%3A%20tcp%20binding%20address
[common-error-2]: https://github.com/NibiruChain/nibiru/issues/1623#common-error-2-spot-tx-hash-not-found:~:text=Test%3A%20%20%20%20%20%20%20%09TestIntegrationTestSuite/TestSuccessfulVoting-,Common%20Error%202%3A%20spot%20tx%20hash%20not%20found,-%2D%2D%2D%20FAIL%3A%20TestIntegrationTestSuite/TestCreatePoolCmd

## Common Error 1: tcp binding address in use

```
--- FAIL: TestIntegrationTestSuite (16.77s)
    --- FAIL: TestIntegrationTestSuite/TestSuccessfulVoting (16.77s)
        network.go:190: acquiring test network lock
        network.go:200: preparing test network...
        network.go:467: starting test network...
        network.go:473: started validator 0
        network.go:473: started validator 1
        app_test.go:46: 
            	Error Trace:	/home/runner/work/nibiru/nibiru/x/oracle/integration/app_test.go:46
            	            				/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.8.4/suite/suite.go:187
            	Error:      	Received unexpected error:
            	            	listen tcp 0.0.0.0:41135: bind: address already in use
            	Test:       	TestIntegrationTestSuite/TestSuccessfulVoting
```

[...]

PR #1652, opened 2023-10-27
test: refactors cli.network suites with 'Integration' to use common function

Unique-Divine:
# Summary

1. Makes our implementation of the convention for skipping network tests with `testing.short` and the `-short` flag consistent across modules. 

2. Makes it so that simulations (`simapp` tests) are filtered out without us needing to use `go list`. As can be seen in `contrib/make/simulation.mk`, the simulation tests use an `-Enabled=true` flag in setup and aren't considered unit tests (using the `-short` flag). Their setup can follow the same convention as the integration tests by calling `testutil.BeforeIntegrationSuite`.

3. ci: In the `unit-test.yml` GitHub action, there's no reason to save a coverage.txt since it's not published or visible from outside the workflow. We're only using the reports from `integration-test.yml`, so it makes sense to use `make test-unit` here and have the action run faster.

- Related to #1623 

# Purpose

1. This change makes the code less repetitive and easier to maintain by having the doc comment on a common function. 

2. Unit tests run in 3-6 minutes now instead of 8+ min. 

3. This will fix the false-alarm CI failures resulting from `cli.Network` tests running when they shouldn't be. For example, https://github.com/NibiruChain/nibiru/actions/runs/6566058545/job/17835937660?pr=1643:

    <details>
      <summary>[Expand example block]</summary>

      ```
      -- FAIL: TestIntegrationTestSuite (16.63s)
          --- FAIL: TestIntegrationTestSuite/TestSuccessfulVoting (16.63s)
              network.go:190: acquiring test network lock
              network.go:200: preparing test network...
              network.go:467: starting test network...
              network.go:473: started validator 0
              app_test.go:46: 
                      Error Trace:	/home/runner/work/nibiru/nibiru/x/oracle/integration/app_test.go:46
                                                  /home/runner/go/pkg/mod/github.com/stretchr/testify@v1.8.4/suite/suite.go:187
                      Error:      	Received unexpected error:
                                      listen tcp 0.0.0.0:35143: bind: address already in use
                      Test:       	TestIntegrationTestSuite/TestSuccessfulVoting
      FAIL
      ```
    </details>


.github/workflows/unit-tests.yml
```diff
@@ -22,17 +22,17 @@ jobs:
           cache: true
 
       - name: Run all unit tests.
-        run: make test-coverage
+        run: make test-unit
```
Unique-Divine, line 25: In the `unit-test.yml` GitHub action, there's no reason to save a coverage.txt since it's not published or visible. We're actually using the reports from `integration-test.yml`. We can use `make test-unit` to have the CI run faster.

contrib/make/test.mk
```diff
@@ -2,11 +2,13 @@
...
 
-PACKAGES_NOSIMULATION = ${shell go list ./... | grep -v simapp}
+.PHONY: test-unit
+test-unit:
+	go test ./... -short
 
-.PHONY: test-coverage
-test-coverage:
-	go test ./... $(PACKAGES_NOSIMULATION) -short \
+.PHONY: test-coverage-unit
```
Unique-Divine, line 9: Using `$PACKAGES_NOSIMULATION` is unnecessary since simulations require a specific flag and we're already using `-short`

x/common/testutil/cases.go
```diff
@@ -18,3 +18,19 @@ func RunFunctionTests(t *testing.T, testCases []FunctionTestCase) {
 		})
 	}
 }
+
+/*
+BeforeIntegrationSuite: Skips a test if the `-short` flag is not used:
```
Unique-Divine, line 23:
```suggestion
BeforeIntegrationSuite: Skips a test if the `-short` flag is used:
```

jgimeno (approved):

PR #1656, opened 2023-10-31
feat(perp): Make the collateral denom a stateful collections.Item

matthiasmatt:
# Description

Create a collateral object to use instead of the quote denom. Collateral is then accessed through Collateral[QuoteDenom]

- Adds `MsgChangeCollateralDenom` to the perp `MsgServer`, gated behind `SudoKeeper.CheckPermissions`

# Purpose

The collateral denom for x/perp needs to be a collections.Item / KeySet to make it configurable.

This is necessary for setting the address for NUSD, which would look like "tf/[contract-addr]/nusd".
This change would also lower the possibility of mistakes arising from hard-coded denominations on all of the bank transfer logic in x/perp.


wasmbinding/bindings/msg.go
```diff
@@ -100,6 +100,11 @@ type SetMarketEnabled struct {
 	Enabled bool   `json:"enabled"`
 }
 
+type UpdateCollateral struct {
+	Denom           string `json:"denom"`
+	ContractAddress string `json:"contract_address"`
+}
+
```
Unique-Divine, line 107: Instead of using the `wasmbinding` folder, it simply add the keeper call registered to the message under the `Admin` field, gated by the `SudoKeeper`

proto/nibiru/perp/v2/genesis.proto
```diff
@@ -23,15 +24,20 @@ message GenesisState {
...
     string volume = 3 [
       (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int",
       (gogoproto.nullable) = false
     ];
+
   }
+  
+  // For testing purpose, we allow the collateral to be set at genesis
+  string collateral_creator = 11;
+  string collateral_subdenom = 12;
```
Unique-Divine, line 40: Let's use a single string for this like `collateral_denom` so that it can be any valid denomination: native, IBC, or token factory.

Unique-Divine (changes requested):
Logic looks good. Here are some suggested changes

Unique-Divine (changes requested):
Final note: All suggested changes have been addressed excepting https://github.com/NibiruChain/nibiru/pull/1656#discussion_r1391796431

proto/nibiru/perp/v2/genesis.proto
Unique-Divine: - Addressed in [0065ae0](https://github.com/NibiruChain/nibiru/pull/1656/commits/0065ae0b57123bdf6e92bf19a6202bc5b6a0ebf9)

wasmbinding/bindings/msg.go
Unique-Divine: Addressed by https://github.com/NibiruChain/nibiru/pull/1656/commits/2dd5807e12f4c1b99a5c9f62226ff8c5a984140d

Unique-Divine (approved):

PR #1658, opened 2023-10-31
feat(oracle): edit oracle params msg

k-yang:
# Description

Creates a `EditOracleParams` message which authorized users/cw contracts can call.

# Purpose

Simplifies the controller contract out of the system.


Unique-Divine (changes requested):
This removed the queries, which you still need. Only the custom ones are what you should delete, not the stargate stuff

Unique-Divine (approved):
LGTM

PR #1660, opened 2023-11-03
chore(wasm): bump wasmd dependency to v0.43.0

k-yang:
# Description

Bumps wasmd to [v0.43.0](https://github.com/CosmWasm/wasmd/blob/v0.43.0/CHANGELOG.md)

# Purpose

Our previous version of wasmd v0.40.2 was old.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.19
 require (
 	cosmossdk.io/errors v1.0.0
 	cosmossdk.io/math v1.1.2
-	github.com/CosmWasm/wasmd v0.40.2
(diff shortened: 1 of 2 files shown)


k-yang:
> It looks like the linter doesn't want the aliased imports. The error comes from here, but it's not actually causing a bug.
> 
> * See https://github.com/CosmWasm/wasmd/blob/v0.43.0/x/wasm/alias.go
> 
> Want to merge it as is, or should we appease the linter?
> 
> The fix is to directly import things from `wasm/keeper` rather than from `wasm`
> 
> image

I can fix it

Issue #1668, opened 2023-11-07
[epic] Smoothness on the x/inflation curve to implement tokenomics more precisely


Unique-Divine:
## Requirements

- [ ] Smaller granularity on changes to the rate of inflation. This should be at least changing every month, but we could set the buckets to a week or even a day since we'll have a continuous function.

 <details>
  <summary>[Expand for image] For context, the current implementation can only 
     handle year-wise step functions distributed on a daily basis. </summary>

   ![image](https://github.com/NibiruChain/nibiru/assets/51418232/b7cb3134-8478-4032-a381-6ff51f7650e9)
  </details>

- [ ] Use order-5 polynomial instead of the (a=initialValue, r=decayRate, c=eventualValue) curve.
- [ ] Bake in a stopping point to automatically toggle off the inflation after passing the "target supply" (1.5B NIBI)

PR #1670, opened 2023-11-08
feat(inflation)!: make inflation follow a polynomial distribution

matthiasmatt:
# Description

Make the inflation follow a polynomial distribution instead of the exponential one.
Inflation period had to be changed to monthly instead of yearly to have better control on the curve.

- Partially closes https://github.com/NibiruChain/nibiru/issues/1668


Unique-Divine (approved):
Tentative pass but several test cases still need to be fixed

PR #1694, opened 2023-12-05
ci: replace `master` with `main`

k-yang:
# Description

Substitute references from `master` branch to `main` branch.

# Purpose

image

## Summary by CodeRabbit

- **Chores**
  - Updated GitHub Actions workflows to reference the new default branch name "main" instead of "master."

- **Documentation**
  - Updated all workflow files to reflect the industry-wide shift in naming conventions for the default branch.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -3,7 +3,7 @@ name: Publish Docker Image
 on:
   push:
     branches:
-      - master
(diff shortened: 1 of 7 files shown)


Unique-Divine (approved):

Unique-Divine:
Note that some of the parts that say main still need to be master if they're in the workflow name

PR #1719, opened 2023-12-18
refactor(test): add is not mandatory interface to action

jgimeno:
# Description

This PR introduces a refactor of the isMandatory boolean argument, replacing it with the IsNotMandatory interface. This change aims to enhance code clarity and flexibility.

# Purpose

Simplified Exception Handling: Previously, isMandatory was predominantly set to true, which often led to redundant checks and less intuitive exception handling. By introducing the IsNotMandatory interface, we make the default path (mandatory) implicit, simplifying the code and focusing exception handling only where it's necessary.

Ease of Modification: Changing an action from mandatory to non-mandatory is now more straightforward. Instead of modifying the boolean value across various function calls, we simply implement the IsNotMandatory method for the relevant action. This approach is more readable and reduces the risk of errors in updating the boolean value in multiple places.

Improved Code Readability: The use of an interface, as opposed to a boolean, makes the intentions of the code clearer. When a developer sees the IsNotMandatory interface, it's immediately apparent that the associated action has optional characteristics, enhancing the understandability of the codebase.

## Summary by CodeRabbit

- **Refactor**
  - Streamlined control flow and error handling across various action and assertion methods by standardizing return values.

- **New Features**
  - Introduced a new non-mandatory interface for actions to enhance testing flexibility.

- **Documentation**
  - Updated CHANGELOG to reflect the introduction of non-mandatory interfaces in testing framework.


Unique-Divine (commented):
Still reading. It's so much cleaner without the `isMandatory` arg

Unique-Divine (approved):

PR #1734, opened 2023-12-28
feat(perp): MsgDonateToPerpFund admin call

Unique-Divine:
# Purpose / Abstract

- Partially completes #


x/perp/v2/keeper/admin.go
```diff
@@ -23,34 +23,42 @@ import (
...
+// - ctx: Blockchain context holding the current state
+// - amount: Amount of micro-NUSD to withdraw.
+// - sender: Admin address registered in x/sudo
+// - to: Recipient address
+func (k admin) WithdrawFromPerpFund(
+	ctx sdk.Context, amount sdkmath.Int, sender, to sdk.AccAddress, denom string,
 ) (err error) {
-	collateral, err := k.Collateral.Get(ctx)
-	if err != nil {
+	if err := k.SudoKeeper.CheckPermissions(sender, ctx); err != nil {
```
jgimeno, line 36: I have some opinions related to checking permissions here. I created some PR here: https://github.com/NibiruChain/nibiru/pull/1751

Unique-Divine: This is a stateful validation, reading from the KV store of another module, so it's not really for the MsgServer

jgimeno: Yeah. It does not change that principle if this call is still in the message server. MsgServer delegates validation of the request to the SudoKeeper (that he really is the one reading the KVStore, not the MsgServer), and if it is correct, it calls the Admin function.

What I don't believe is in the admin function to validate the permissions. It is a thin line and a matter of convention, but I can show you some benefits of this slight separation.

proto/nibiru/perp/v2/tx.proto
```diff
@@ -429,3 +434,19 @@ message MsgShiftSwapInvariant {
...
+// WithdrawFromPerpFund: gRPC tx msg for changing the swap invariant.
+// Admin-only.
+message MsgWithdrawFromPerpFund {
+  string sender = 1;
+  string amount = 2 [
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int",
+    (gogoproto.nullable) = false
+  ];
+  string denom = 3;
+  string to_addr = 4;
```
k-yang, line 449: (nit) it might be safer to just withdraw to whoever sends the message, i.e. the message `sender`.

I can foresee possible human error in the future where somebody sends the message but accidentally sets the `to_addr` different from the message signer/sender.

x/perp/v2/keeper/admin.go
Unique-Divine: Checking a whitelist is not a message validation. If the keeper method doesn't have the permission check, other modules that import the methods can sidestep the sudo, defeating the purpose for the check. 

It's not just a difference by convention. Switching it to `MsgServer` changes the security of the keeper method itself.

proto/nibiru/perp/v2/tx.proto
```diff
@@ -429,3 +434,19 @@ message MsgShiftSwapInvariant {
...
 
 message MsgShiftSwapInvariantResponse {}
+
+// -------------------------- WithdrawFromPerpFund --------------------------
+
+// WithdrawFromPerpFund: gRPC tx msg for changing the swap invariant.
+// Admin-only.
+message MsgWithdrawFromPerpFund {
+  string sender = 1;
+  string amount = 2 [
```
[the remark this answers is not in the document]
Unique-Divine: Yeah, there's reasoning behind this. Usually, we'll leave the denom blank. For this message, blank -> default -> use the collateral denom `tf/..../nusd`, which is pretty long and usually what we intend to use. 

We'd only use `denom` if we wanted something besides the default collateral that ended up in the Perp Fund.

proto/nibiru/perp/v2/tx.proto
Unique-Divine: I think so too. The problem we might encounter here is that the claim goes to a smart contract that doesn't expose a function to withdraw its balance. 

`sudo_contract_calls_msg -> withdraw_to_contract_addr -> No fn to withdraw -> funds are lost`

I'm trying to avoid this ^ deadlock.

Adding an explicit `to_addr` makes it less error prone in a different way, by making the caller pass in the location they want the funds to end up. 

Another option is to make that field nullable, so that empty means, `to_addr := sender`. Would you prefer that?

x/perp/v2/keeper/admin.go
jgimeno: The message field validation should be done by default out of the message server as long as you implement ValidateBasic(). Validating any field in the message server is duplicating computing cycles. What, then, is the use case of the MsgServer if we remove the ValidateBasic?

There are two paradigms here: the MsgServer as orchestrator of keepers (my ideal) or the Keepers as holders of all the logic and being pure defensive (your ideal; this removes the need for MsgServer in my opinion as it becomes only a wrapper of the keeper).

It is a matter of convention from the moment you agree on what it means to share some Keeper method. Is there any use case at this stage for CloseMarket to be public in the keeper? Try to avoid the what-if thing. 

I am open to anything as long as it makes sense, and we will follow the convention from now on.

CHANGELOG.md
```diff
@@ -66,6 +66,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * [#1573](https://github.com/NibiruChain/nibiru/pull/1573) - feat(perp): Close markets and compute settlement price
 * [#1705](https://github.com/NibiruChain/nibiru/pull/1705) - feat(perp): Add oracle pair to market object
 * [#1718](https://github.com/NibiruChain/nibiru/pull/1718) - fix: fees does not require additional funds
+* [#1734](https://github.com/NibiruChain/nibiru/pull/1734) - feat(perp): MsgDonateToPerpFund
```
jgimeno, line 69: A little bit of a more descriptive message like the one in the title would be nice. You will appreciate it when you search for this in the changelog in four years!

proto/nibiru/perp/v2/tx.proto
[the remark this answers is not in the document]
k-yang: Gotcha, makes sense.

proto/nibiru/perp/v2/tx.proto
k-yang: Wow, I didn't even consider that scenario. In that case, then I think it's better to be explicit. 

Thinking about it more, in the general case, we'll likely be sending these `WithdrawFromPerpFund` messages from the CW3 multisig. Having the ability to target a recipient with the message would be more efficient, instead of withdrawing to the CW3 multisig and then doing a bank transfer.

A more sophisticated design could be to only run the message if the `to_addr` is part of the `sudo` sudoers list. It would prevent the funds from going to a CosmWasm smart contract that doesn't have the withdraw functionality, because we would be the ones whitelisting the possible recipients. But that could be addressed in a separate PR. What's implemented looks good to me.

k-yang (approved):

PR #1751, opened 2023-12-28
docs: add adr-1 proposal for msg server and keeper consistency.

jgimeno:
# Purpose / Abstract

## Summary by CodeRabbit

- **Documentation**
	- Introduced a proposal for refactoring core components to enhance separation of concerns and testing framework within the Nibiru Chain documentation.
	- Updated the ADR template to include structured sections for documenting architectural decisions.
	- Added an ADR outlining the separation of concerns between `MsgServer` and `Keeper` in the Nibiru Chain.
	- Introduced an ADR framework for managing significant software design choices in the documentation.


Unique-Divine (commented):
Looks good. Just going to rename the file path and add some text improvements, then merge this.

Unique-Divine (changes requested):
One thing to note here is that `x/sudo` checks are actually stateful validation, which would still go in the `Keeper`. They're not a msg validation for the `MsgServer`. 

The ADR should probably propose a format for that or advocate for some convention like the original proposal from Frojdi

matthiasmatt:
LGTM, just want everyone here to pinky promise that we will never do a binding from msg to keeper ever again. Any binding (like wasm for example) needs to go to msg server and not touch the keeper, otherwise we can lose the msg validation from the msg server.

Unique-Divine (approved):

PR #1800, opened 2024-02-18
fix(localnet): script seems broken

Unique-Divine:
# Purpose / Abstract

Working on #1799 and the `loclanet.sh` script keeps failing. In addition, `nibid init` is taking an absurdly long amount of time to run.

## Summary by CodeRabbit

- **Bug Fixes**
	- Modified the inflation allocation function to correctly return an error when expected, enhancing the reliability of inflation handling.


Unique-Divine:
Fixed in another PR

PR #1806, opened 2024-02-21
ci: fix chaosnet build

k-yang:
# Purpose / Abstract

https://github.com/NibiruChain/nibiru/commit/688ac5df29f20213275f298a9a4ed6eb3fc5ca01 added feature flags, but forgot to copy those scripts to the Dockerfile, so we end up with `./chaosnet.sh: line 48: ./feat-perp.sh: No such file or directory` in [build workflows](https://github.com/NibiruChain/nibiru/actions/runs/7981567278/job/21793501254).

Also removes some unnecessary variables from the script.

## Summary by CodeRabbit

- **New Features**
	- Enhanced Docker support by adding additional script capabilities.

- **Refactor**
	- Simplified script sourcing in network management scripts by adjusting path references.
	- Streamlined feature flag initialization in network setup scripts.


Unique-Divine (changes requested):
The curr_dir part is actually needed to resolve different paths depending on where you call the script from. It doesn't resolve local paths on its own when using source ./

Unique-Divine (approved):

PR #1874, opened 2024-05-14
fix: remove all stringer for proto types

matthiasmatt:
As per https://docs.cosmos.network/v0.50/build/migrations/upgrading#stringer, remove the stringer from the gogo proto.

For TD denoms, I had to create a pretty print for the "tf/contract_address/denom" logic.


x/tokenfactory/types/state.go
Unique-Divine, line 1: I think it'd make more sense to put the implementation inside `Denom` and call `DenomStr` everywhere rather than `PrettyString` because it's not so much a pretty print as it is the default string representation.

Then you'd end up with 
```go
func (tfd TFDenom) DenomStr() string {
	return DenomStr(tfd.Denom())
}

func (tfd TFDenom) Denom() DenomStr {
	return fmt.Sprintf("tf/%s/%s", tfd.Creator, tfd.Subdenom)
}
```

Unique-Divine (changes requested):

x/tokenfactory/types/state.go
matthiasmatt: Why not Denom directly? what does denomstr bring?

Unique-Divine (approved):

PR #1892, opened 2024-05-23
feat: update to cosmos sdk v0.50.6

matthiasmatt:
# Purpose / Abstract

- Closes #AAA

## Things to test:

- [ ] app/upgrades/v1_3_0/constants.go -> Test that this does not panic due to the introduced `sdkCtx := sdk.UnwrapSDKContext(ctx)`


matthiasmatt:
image

PR #1907, opened 2024-06-03
test(evm): grpc_query full coverage

onikonychev:
## Summary by CodeRabbit

- **New Features**
  - Enhanced test coverage for `grpc_query` functionality in the EVM module.
  - Added functions for tracing and executing Nibi and ERC-20 transfers.
  - Introduced new test cases for querying Ethereum calls, balances, base fees, and gas estimation for EVM calls.

- **Bug Fixes**
  - Corrected minor documentation errors in test files.

- **Refactor**
  - Improved code readability by removing unnecessary parentheses and updating function signatures.


x/evm/keeper/grpc_query_test.go
```diff
@@ -447,38 +467,518 @@ func (s *KeeperSuite) TestQueryParams() {
...
+			func(deps *evmtest.TestDeps) (req In, wantResp Out) {
+				txMsg, predecessors := s.ExecuteERC20Transfer(deps)
+
+				req = &evm.QueryTraceTxRequest{
+					Msg:          txMsg,
+					Predecessors: predecessors,
+				}
+				wantResp = TraceERC20Transfer()
+				return req, wantResp
+			},
```
Unique-Divine, line 785: I would write the `ExecuteERC20Transfer` function as taking arguments `(deps, s)` rather than including it as a function on the suite. That will make it easier to pull out the testutil to the `evmtest` import if we want to leverage it in other places (like `nibiru/eth` or `nibiru/app`) 

Perhaps the `ExecuteNibiTransfer` fits this criterion as well.

x/evm/keeper/grpc_query_test.go
```diff
@@ -447,38 +467,518 @@ func (s *KeeperSuite) TestQueryParams() {
...
+	gethSigner := deps.Sender.GethSigner(deps.Chain.EvmKeeper.EthChainID(deps.Ctx))
+	keyringSigner := deps.Sender.KeyringSigner
+	return txMsg, txMsg.Sign(gethSigner, keyringSigner)
+}
+
+func GenerateEthAddress() gethcommon.Address {
+	privateKey, _ := crypto.GenerateKey()
+	publicKey := privateKey.Public()
+	publicKeyECDSA, _ := publicKey.(*ecdsa.PublicKey)
+	return crypto.PubkeyToAddress(*publicKeyECDSA)
```
Unique-Divine, line 983: This should be replaced with the equivalent util fn from `evmtest` if possible. It looks almost the same as `evmtest.NewEthAccInfo` from eth.go. 

If these functions are not the same, can you add it to `evmtest` and put a function comment on the difference?

x/evm/keeper/grpc_query_test.go
```diff
@@ -1,26 +1,53 @@
...
 	"github.com/NibiruChain/nibiru/x/common/testutil/testapp"
 	"github.com/NibiruChain/nibiru/x/evm"
 	"github.com/NibiruChain/nibiru/x/evm/evmtest"
 )
 
 func InvalidEthAddr() string { return "0x0000" }
 
+func TraceNibiTransfer() string {
+	return `{
+	  "gas": 21000,
```
Unique-Divine, line 28: Why is it this number specifically? Rather, can you replace this with the causal constant from "gethparams" and use `fmt.Sprintf`?

Unique-Divine (changes requested):
Looks great overall. Left a few quick review comments

Unique-Divine (approved):

Unique-Divine (approved):

PR #1966, opened 2024-07-13
refactor(evm): clean up AnteHandler setup

k-yang:
# Purpose / Abstract

- Remove `ExtensionOptionDynamicFeeTx`. It's an Evmos-specific construct that tries to add `baseFee` and `priorityFee` to regular Cosmos Module Txs. We don't need it, as `MsgEthereumTx` already has `baseFee` and `priorityFee` in EVM-module txs.
- Remove `ExtensionOptionsWeb3Tx`. It's a legacy EIP-712 method of signing txs, which Ethermint used to have but has been removed in Evmos. It's simply dead code in our codebase, so I removed it.
- Simplifies switching between EVM ante handlers and non-EVM antehandlers through detecting the `ExtensionOptionsEthereumTx` tx extension.


app/ante.go
```diff
@@ -10,64 +10,51 @@ import (
...
-		hasExt, typeUrl := TxHasExtensions(tx)
-		if hasExt && typeUrl != "" {
-			anteHandler = AnteHandlerExtendedTx(typeUrl, keepers, opts)
-			return anteHandler(ctx, tx, sim)
+		txWithExtensions, ok := tx.(authante.HasExtensionOptionsTx)
+		if ok {
+			opts := txWithExtensions.GetExtensionOptions()
+			if len(opts) > 0 {
+				switch typeURL := opts[0].GetTypeUrl(); typeURL {
+				case "/ethermint.evm.v1.ExtensionOptionsEthereumTx":
```
Unique-Divine, line 36: Wrong type url here

app/ante.go
Unique-Divine, line 1: It looks like this is deprecating LegacyTx, but that's still in use. I believe it's the main one used by ethers and other clients

Unique-Divine (changes requested):

eth/codec.go
Unique-Divine, line 1: Why does removing dynamic fee tx necessitate removing Web3 tx?

Unique-Divine (changes requested):

app/ante.go
k-yang: Fixed.

eth/codec.go
k-yang: They're unrelated to each other, but both are unnecessary for our EVM implementation.

Ethermint had `ExtensionOptionsWeb3Tx` to support legacy EIP-712 signature extensions, but those were deprecated and we can just validate the signature on the Cosmos tx payload.


`ExtensionOptionDynamicFeeTx` was to add baseFee and priorityFee to Cosmos native module txs (i.e. non-EVM txs), but we don't have that in scope right now.

Unique-Divine (approved):

PR #1969, opened 2024-07-22
spike(evm): use cosmos address derivation for ethsecp256k1

k-yang:
# Purpose / Abstract

Cosmos uses RIPEMD160(SHA256()), whereas Ethereum uses Keccak-256.


k-yang:
Not proceeding with this anymore

PR #2004, opened 2024-08-15
refactor(evm)!: replace `HexAddr` with `EIP55Addr`

k-yang:
# Context

Please see https://eips.ethereum.org/EIPS/eip-55 for the EIP-55 spec.

# Purpose / Abstract

Instead of having a `HexAddr` type that "sublasses" the primitive `string` type, we introduce an `EIP55Addr` type which "subclasses" `gethcommon.Address`.  `EIP55Addr` is a stricter type than `HexAddr`, because only valid representations of `gethcommon.Address` are valid `EIP55Addr`s. We also still inherit from `sdk.CustomProtobufType`, keeping the benefit of using `EIP55Addr` as a custom proto field and allowing it to be serialized into collections state, but avoiding the pitfall described below:

`HexAddr` was dangerous because one could do `addr := eth.HexAddr("KEVIN")` and bypass the validation checks in `HexAddr.Valid()`. `EIP55Addr` requires the creator to use `NewEIP55AddrFromStr` or pass a valid `gethcommon.Address` to `EIP55Addr{Address: addr}`, and `gethcommon.Address` is a canonical 20-byte array, so it can never have alternative representations (e.g. mixed-case alphanumeric characters, or alternative lengths).

The underlying `gethcommon.Address` can be accessed via `eip55Addr.Address`, providing a field lookup instead of re-calculating the entire `gethcommon.Address` from a string. We also inherit all of the `gethcommon.Address` methods, like `String()` and `Bytes()`, which are helpful for proto marshalling.

The PR also refactors tests to use the new `EIP55Addr` type.

BREAKING CHANGE: changes underlying proto types from `HexAddr` to `EIP55Addr`.


Unique-Divine:
You can bypass the validation of any type by casting in the way you've described. 

Writing `eth.HexAddr("KEVIN")` is just as dangerous as writing `gethcommon.Address([]byte("KEVIN"))`

Unique-Divine (approved):
Nice, this accomplishes what we wanted 🙌

k-yang:
> > `HexAddr` was dangerous because one could do `addr := eth.HexAddr("KEVIN")` and bypass the validation checks in `HexAddr.Valid()`. `EIP55Addr` requires the creator to use `NewEIP55AddrFromStr` or pass a valid `gethcommon.Address` to `EIP55Addr{Address: addr}`, and `gethcommon.Address` is a canonical 20-byte array, so it can never have alternative representations (e.g. mixed-case alphanumeric characters, or alternative lengths).
> 
> There isn't actually a tradeoff in safety here. You can bypass the validation of any type by casting in the way you've described.
> 
> Writing `eth.HexAddr("KEVIN")` is just as dangerous as writing `gethcommon.Address([]byte("KEVIN"))`

For documentation purposes:
`gethcommon.Address([]byte(...)) is safer than `eth.HexAddr(...)` because gethcommon.Address forces a 20-byte array, whereas `eth.HexAddr()` can be of any string length, so `gethcommon.Address([]byte("KEVIN"))` will fail compilation but `eth.HexAddr("KEVIN")` will pass compilation, but it'll fail `eth.HexAddr.Valid()`.

Also, any 20-byte array is a valid Ethereum address, so `gethcommon.Address([]byte("KEVINKEVINKEVINKEVIN"))` compiles and is usable as a valid address because the letters get parsed as UTF-8 and converted into bytes, whereas `eth.HexAddr("KEVINKEVINKEVINKEVIN")` passes compilation but is unsafe because the string is not hex-encoded.

PR #2056, opened 2024-10-01
feat(evm): add oracle precompile

### LINKED ISSUE #2051: [evm] Minimalist precompile for x/oracle
## The Ask

Data feeds published by validator operators in the Oracle Module are currently
inaccessible through the EVM. We can implement a precompile that exposes the "Query/ExchangeRate" RPC method.

## Future Considerations

The x/oracle module uses a lot of terminology specific to price data. It is common for
oracles to provide data like pseudo-random numbers, so we should survey for how
much demand that would have from builders of applications on Nibiru.

## Note on Composability

> [!TIP]
> Note that this is technically covered by this ticket because we can deploy a Wasm interface for x/oracle and have the EVM call it:
> - #2048

k-yang:
# Purpose / Abstract

- Closes #2051

## Summary by CodeRabbit

- **New Features**
	- Introduced an Oracle contract for querying exchange rates within the Ethereum Virtual Machine.
	- Added support for Ethereum transaction types and enhanced inflation parameters.
	- Implemented an in-memory EventBus for real-time event distribution.
	- Added a new custom precompile for the Oracle contract.

- **Bug Fixes**
	- Resolved issues with keyring import functions and FunToken state marshalling.

- **Documentation**
	- Updated the changelog to reflect new features, improvements, and bug fixes.


x/evm/embeds/contracts/IOracle.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.19;
+
+/// @dev Oracle interface for querying exchange rates
+interface IOracle {
+  /// @dev queryExchangeRate queries the exchange rate for a given pair
+  /// @param pair the pair to query
```
Unique-Divine, line 7:
```suggestion
/// @notice Oracle interface for querying exchange rates
interface IOracle {
  /// @notice Queries the exchange rate for a given pair
  /// @param pair The asset pair to query. For example, "ubtc:uusd" is the 
  /// USD price of BTC and "unibi:uusd" is the USD price of NIBI.
  /// @return The exchange rate (a decimal value) as a string.
  /// @dev This function is view-only and does not modify state.
```

I based this on the guidelines [written here](https://github.com/NibiruChain/nibiru/tree/main/x/evm/embeds#precompile-solidity-documentation).

x/evm/embeds/contracts/IOracle.sol
```diff
@@ -0,0 +1,13 @@
...
+  function queryExchangeRate(string memory pair) external returns (string memory);
```
Unique-Divine, line 8:
```suggestion
  function queryExchangeRate(string memory pair) external view returns (string memory);
```

We should add the `view` keyword here since the contract doesn't modify state.

x/evm/precompile/oracle.go
```diff
@@ -0,0 +1,130 @@
...
+		return nil, err
+	}
+
+	switch OracleMethod(method.Name) {
+	case OracleMethod_QueryExchangeRate:
+		// TODO: UD-DEBUG: Test that calling non-method on the right address does
+		// nothing.
+		bz, err = p.queryExchangeRate(ctx, method, args, readonly)
+	default:
+		// TODO: UD-DEBUG: test invalid method called
```
Unique-Divine, line 73:
```suggestion
		bz, err = p.queryExchangeRate(ctx, method, args, readonly)
	default:
```

I later discovered that both scenarios described by those TODO comments are impossible because of the Input.Unpack logic inside of `DecomposeInput`. The `switch` block is relatively safe, so long as we have a case for each method defined in the precompile's ABI.

x/evm/precompile/oracle.go
```diff
@@ -0,0 +1,130 @@
...
+	err error,
+) {
+	if len(args) != 1 {
+		err = fmt.Errorf("expected 3 arguments but got %d", len(args))
+		return
+	}
+
+	pair, ok := args[0].(string)
+	if !ok {
+		err = fmt.Errorf("type validation for failed for (address erc20) argument")
```
Unique-Divine, line 125:
```suggestion
		err = ErrArgTypeValidation("string pair", args[0])
```

I added a helper function to make the error messages more consistent between precompiles.

Unique-Divine (changes requested):
LGTM, added a few small changes

PR #2094, opened 2024-10-26
fix(evm): StateDB multistore cache for precompile reversion and a safer Nibiru bank keeper that respects the EVM

Unique-Divine:
# Purpose / Abstract

fix(evm): Following from the changes in #2086, this pull request implements a critical security
fix

## Multistore copy to safely revert precompile changes outside EVM

First, we add new `JournalChange` struct that saves a deep copy of the
state multi store before each state-modifying, Nibiru-specific precompiled
contract is called (`OnRunStart`). Additionally, we commit the `StateDB` there
as well. This guarantees that the non-EVM and EVM state will be in sync even
if there are complex, multi-step Ethereum transactions, such as in the case of
an EthereumTx that influences the `StateDB`, then calls a precompile that also
changes non-EVM state, and then EVM reverts inside of a try-catch.

## Related Tickets

- https://github.com/code-423n4/2024-10-nibiru-zenith/issues/43
- https://github.com/code-423n4/2024-10-nibiru-zenith/issues/31
- https://github.com/NibiruChain/pm/issues/25

## Release Notes

- **New Features**
	- Introduced `JournalChange` struct for state consistency during contract calls.
	- Added `IsPrecompile` method for checking precompiled contracts.
	- Enhanced transaction handling with `IncrementWasmCounterWithExecuteMulti` and new VM call support.
	- Added `NewStateDB` method for creating state database instances.
	- Introduced `PrecompileCalled` struct for managing state changes from precompiles.

- **Improvements**
	- Enhanced error handling and validation in various methods, especially for Ethereum transaction processing.
	- Streamlined context management in `StateDB` for better state handling.
	- Improved clarity in changelog entries and structured updates under "Nibiru EVM".

- **Bug Fixes**
	- Resolved state consistency issues and improved gas fee handling in ERC20 contract execution.

- **Documentation**
	- Updated changelog format for clarity and organization of changes.


app/keepers.go
```diff
@@ -605,7 +619,7 @@ func (app *NibiruApp) initAppModules(
 		),
 		auth.NewAppModule(appCodec, app.AccountKeeper, authsims.RandomGenesisAccounts, app.GetSubspace(authtypes.ModuleName)),
 		vesting.NewAppModule(app.AccountKeeper, app.BankKeeper),
-		bank.NewAppModule(appCodec, app.BankKeeper, app.AccountKeeper, app.GetSubspace(banktypes.ModuleName)),
+		bank.NewAppModule(appCodec, app.bankBaseKeeper, app.AccountKeeper, app.GetSubspace(banktypes.ModuleName)),
```
Unique-Divine, line 622: This line's required to appease the module manager when it does a type assertion that the object of type `bankkeeper.Keeper` (an interface) is actually the `bankkeeper.BaseKeeper` struct. I didn't see a simple way to force x/bank to use the `NibiruBankKeeper` struct without editing the Bank module directly, so I opted for what's written here, which is much simpler.

go.mod
```diff
@@ -244,6 +244,9 @@ require (
 replace (
 	cosmossdk.io/api => cosmossdk.io/api v0.3.1
 
+	github.com/CosmWasm/wasmd => github.com/NibiruChain/wasmd v0.44.0-nibiru
+	github.com/cosmos/cosmos-sdk => github.com/NibiruChain/cosmos-sdk v0.47.11-nibiru
```
Unique-Divine, line 248: Small amendments to our current pinned versions. These are needed to create deep copies of the multistore in statedb/journal.go

x/evm/statedb/config.go
Unique-Divine, line 1: Comments like these were conflicting with Godoc's package-level documentation. Since I added docs, I needed to move these lines

x/evm/keeper/erc20.go
Unique-Divine, line 1: This was here because of the repeated calls during gas estimation, not for EVM reversions

x/evm/precompile/funtoken.go
```diff
@@ -72,8 +72,7 @@ func (p precompileFunToken) Run(
 	if err != nil {
 		return nil, err
 	}
-	// Dirty journal entries in `StateDB` must be committed
-	return bz, start.StateDB.Commit()
```
Unique-Divine, line 76: This will be dictated by `ApplyEvmMsg` at the end of the Ethereum tx.

x/evm/precompile/test/export.go
```diff
@@ -308,9 +312,68 @@ func IncrementWasmCounterWithExecuteMulti(
...
 	ethTxResp, evmObj, err := deps.EvmKeeper.CallContractWithInput(
-		deps.Ctx, deps.Sender.EthAddr, &precompile.PrecompileAddr_Wasm, true, input,
+		deps.Ctx, deps.Sender.EthAddr, &precompile.PrecompileAddr_Wasm, finalizeTx, input,
 	)
 	s.Require().NoError(err)
 	s.Require().NotEmpty(ethTxResp.Ret)
 	return evmObj
 }
+
+func IncrementWasmCounterWithExecuteMultiViaVMCall(
```
Unique-Divine, line 322: Uses *vm.EVM.Call instead of a full tx so that intermediate states can be tested

x/evm/statedb/debug.go
Unique-Divine, line 1: I've moved all of the test utils out of the normal body of functions for clarity

x/evm/statedb/journal.go
```diff
@@ -335,3 +318,46 @@ func (ch accessListAddSlotChange) Revert(s *StateDB) {
...
+	Events     sdk.Events
+}
+
+var _ JournalChange = PrecompileCalled{}
+
+// Revert rolls back the [StateDB] cache context to the state it was in prior to
+// the precompile call. Modifications to this cache context are pushed to the
+// commit context (s.evmTxCtx) when [StateDB.Commit] is executed.
+func (ch PrecompileCalled) Revert(s *StateDB) {
+	s.cacheCtx = s.cacheCtx.WithMultiStore(ch.MultiStore)
```
k-yang, line 351: Note to future self: since we only create a `cacheCtx` whenever a precompile is called, then reverting a precompile call can just get rid of the `cacheCtx` and reset the `writeToCommitCtxFromCacheCtx` function to `nil`. That would be a cleaner way to revert a `PrecompileCalled` journal entry rather than deep copy the multistore and save it in a snapshot, but I'd have to test that implementation out.

x/evm/statedb/access_list.go
```diff
@@ -1,3 +1,5 @@
+package statedb
```
k-yang, line 1: Is there a linter that moves package declarations above file-level comments?

go.mod
k-yang: Luckily Cosmos SDK v0.50 implements deep copy for us.

x/evm/keeper/statedb.go
```diff
@@ -17,6 +17,13 @@ import (
 
 var _ statedb.Keeper = &Keeper{}
 
+func (k *Keeper) NewStateDB(
```
k-yang, line 20: I would get rid of this function because it's only a one-line replacement, we use `statedb.New` in other parts of the codebase, and it keeps EVMKeeper more decoupled from StateDB.

k-yang (approved):

x/evm/statedb/journal.go
Unique-Divine: Opening as an improvement ticket

x/evm/keeper/statedb.go
Unique-Divine: The rationale's not as obvious with the next change separated out onto #2095.

For context, this is the full function.  
```go
func (evmKeeper *Keeper) NewStateDB(
	ctx sdk.Context, txConfig statedb.TxConfig,
) *statedb.StateDB {
	stateDB := statedb.New(ctx, evmKeeper, txConfig)
	evmKeeper.Bank.ResetStateDB(stateDB)
	return stateDB
}
```

This constructor's meant to be used any time the EVM keeper generates a StateDB to make sure that its bank keeper reference is always pointing to the current `StateDB` for any EVM transaction.

PR #2095, opened 2024-10-29
fix(evm): Use a NibiruBankKeeper to automatically record ether balance changes on the StateDB journal

Unique-Divine:
# Purpose / Abstract

- Fixes https://github.com/code-423n4/2024-10-nibiru-zenith/issues/31

fix(evm): This change records NIBI (ether) transfers on the `StateDB` during
precompiled contract calls using the `NibiruBankKeeper`, which is a struct
extension of the `bankkeeper.BaseKeeper` that is used throughout Nibiru. The
`NibiruBankKeeper` holds a reference to the current EVM `StateDB` and records
balance changes in wei as journal changes automatically. 

This guarantees that commits and reversions of the `StateDB` do not misalign 
with the state of the Bank module. This code change uses the `NibiruBankKeeper` 
on all modules that depend on x/bank, such as the EVM and Wasm modules. 

```go
type NibiruBankKeeper struct {
	bankkeeper.BaseKeeper
	StateDB                  *statedb.StateDB
	balanceChangesForStateDB uint64
}
```

Each "Send*" method from the bank keeper has an override function that keeps the
the keeper consistent with the Bank when ether is transferred. 

Note however that because calls of `ApplyEvmMsg` initialize new
`*statedb.StateDB` instances even if the calls are for queries, we need to make
sure that `NibiruBankKeeper` references the currnet precompile run's `StateDB`
before any potential use of the bank keeper.

```go
// from x/evm/precompile/wasm.go
p.Bank.StateDB = startResult.StateDB
switch PrecompileMethod(startResult.Method.Name) {
```

## Cosmos-SDK Diff

See [the git diff between the updated Cosmos-SDK dependency
(NibiruChain/cosmos-sdk tag v0.47.11-nibiru.2)
](https://github.com/cosmos/cosmos-sdk/compare/v0.47.11...NibiruChain:cosmos-sdk:v0.47.11-nibiru.2.).
This adds deep copy for the multistore and removes an interface assertion that
makes it possible to plug in the `NibiruBankKeeper` as the bank keeper across
every module.

## Related

- https://github.com/NibiruChain/nibiru/pull/2094
- https://github.com/code-423n4/2024-10-nibiru-zenith/issues/43
- https://github.com/NibiruChain/pm/issues/25

## Release Notes

- **New Features**
	- Introduced fun token creation fee validation and WASM multi-message execution validation.
	- Added support for Foundry and a template for Nibiru EVM development.

- **Bug Fixes**
	- Enhanced handling of ERC20 transfers to address false success values.
	- Improved gas price calculations and state consistency in precompile execution.

- **Refactor**
	- Refactored code to improve clarity and remove outdated comments.
	- Updated bank handling logic to streamline operations and enhance state management.

- **Tests**
	- Added new tests to check for race conditions and improved error handling in existing tests.
	- Updated test cases to utilize a new method for state database instantiation.

These updates enhance the functionality and reliability of the Nibiru EVM, preparing for an upcoming audit.


app/ante/gas_wanted_test.go
```diff
@@ -83,7 +83,7 @@ func (s *AnteTestSuite) TestGasWantedDecorator() {
 	for _, tc := range testCases {
 		s.Run(tc.name, func() {
 			deps := evmtest.NewTestDeps()
-			stateDB := deps.StateDB()
+			stateDB := deps.NewStateDB()
```
Unique-Divine, line 86: The function is renamed from `StateDB` to `NewStateDB` everywhere to reflect that fact that it's not a getter for an existing `StateDB` and instead generates and returns a new pointer for a `StateDB` with a fresh journal

x/evm/embeds/contracts/TestNativeSendThenPrecompileSend.sol
Unique-Divine, line 1: We export an instance of the precompiled contract preset to its address as a constant in the Solidity code for convenience. We still need to publish these contracts on `npm` ([ts-sdk#376](https://github.com/NibiruChain/ts-sdk/issues/376)) after the API becomes more solidified. 

```solidity
interface IFunToken {
    /// ...
}

address constant FUNTOKEN_PRECOMPILE_ADDRESS = 0x0000000000000000000000000000000000000800;

IFunToken constant FUNTOKEN_PRECOMPILE = IFunToken(FUNTOKEN_PRECOMPILE_ADDRESS);
```

This makes it possible to use the functions with strong typing via the imported `FUNTOKEN_PRECOMPILE` variable.

app/keepers.go
```diff
@@ -367,16 +369,17 @@ func (app *NibiruApp) InitKeepers(
...
 
-	app.EvmKeeper = evmkeeper.NewKeeper(
+	evmKeeper := evmkeeper.NewKeeper(
 		appCodec,
 		keys[evm.StoreKey],
 		tkeys[evm.TransientKey],
 		authtypes.NewModuleAddress(govtypes.ModuleName),
 		app.AccountKeeper,
-		app.BankKeeper,
+		nibiruBankKeeper,
```
Unique-Divine, line 378: Written this way to make sure the EVM keeper and its constructor have the stronger typed `NibiruBankKeeper` rather than an a `bankkeeper.Keeper` interface

x/evm/keeper/bank_extension.go
```diff
@@ -0,0 +1,166 @@
...
+	bk.StateDB.SetBalanceWei(eth.NibiruAddrToEthAddr(acc), balanceWei)
+}
+
+func findEtherBalanceChangeFromCoins(coins sdk.Coins) (found bool) {
+	for _, c := range coins {
+		if c.Denom == evm.EVMBankDenom {
+			return true
+		}
+	}
+	return false
```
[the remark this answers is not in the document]
Unique-Divine: I would agree with that but looked into the assumptions in the method and found them unsafe. I'll walk through that here.

cosmos-sdk@v0.47.11/types/coin.go:
```go
// AmountOf returns the amount of a denom from coins
func (coins Coins) AmountOf(denom string) Int {
	mustValidateDenom(denom)
	return coins.AmountOfNoDenomValidation(denom)
}

// AmountOfNoDenomValidation returns the amount of a denom from coins
// without validating the denomination.
func (coins Coins) AmountOfNoDenomValidation(denom string) Int {
	if ok, c := coins.Find(denom); ok {
		return c.Amount
	}
	return ZeroInt()
}
```

This seems harmless at first glance until you read what the `Coins.Find` function says. `Coin.Find` assumes that the coins are sorted and does a binary search to go through the slice and find coin with the denom specified. That's problematic because it assumes   the sequence of coins is sorted, which can cause strange behavior and panics inside the binary search.

We could forcibly validate the `Coins` argument to make sure it's sorted using `Coins.Sorted` or check the validation error value from `Coins.ValidateBasic`, but both of these activities would end up iterating through the entire slice anyway. Thus, I'd argue it's better in terms of performance and safety to simply iterate through the slice to find the desired denom rather than using binary search or anything that could panic.

Put simply, between these options:
1. basic for loop to find denom  
2. `Coins.Sort` -> `Coins.Find` binary search for denom  
3. `if err := Coins.ValidateBasic(); err == nil` -> `Coins.Find`  
  else do (2) 

I think option (1) is the best. 

```go
// Find returns true and coin if the denom exists in coins. Otherwise it returns false
// and a zero coin. Uses binary search.
// CONTRACT: coins must be valid (sorted).
func (coins Coins) Find(denom string) (bool, Coin) {
	switch len(coins) {
	case 0:
		return false, Coin{}

	case 1:
		coin := coins[0]
		if coin.Denom == denom {
			return true, coin
		}
		return false, Coin{}

	default:
		midIdx := len(coins) / 2 // 2:1, 3:1, 4:2
		coin := coins[midIdx]
		switch {
		case denom < coin.Denom:
			return coins[:midIdx].Find(denom)
		case denom == coin.Denom:
			return true, coin
		default:
			return coins[midIdx+1:].Find(denom)
		}
	}
}
```

x/evm/keeper/funtoken_from_coin_test.go
```diff
@@ -304,60 +305,86 @@ func (s *FunTokenFromCoinSuite) TestNativeSendThenPrecompileSend() {
...
 
 	// Alice hex and Alice bech32 is the same address in different representation,
 	// so funds are expected to be available in Alice's bank wallet
 	alice := evmtest.NewEthPrivAcc()
+	evmtest.FunTokenBalanceAssert{
+		FunToken:     funtoken,
+		Account:      alice.EthAddr,
+		BalanceBank:  big.NewInt(0),
+		BalanceERC20: big.NewInt(0),
+	}.Assert(s.T(), deps)
```
Unique-Divine, line 348: Much easier to read IMO

x/evm/keeper/statedb.go
```diff
@@ -79,27 +72,28 @@ func (k *Keeper) ForEachStorage(
 func (k *Keeper) SetAccBalance(
 	ctx sdk.Context, addr gethcommon.Address, amountEvmDenom *big.Int,
 ) error {
-	nativeAddr := sdk.AccAddress(addr.Bytes())
-	balance := k.bankKeeper.GetBalance(ctx, nativeAddr, evm.EVMBankDenom).Amount.BigInt()
+	addrBech32 := eth.EthAddrToNibiruAddr(addr)
+	balance := k.Bank.GetBalance(ctx, addrBech32, evm.EVMBankDenom).Amount.BigInt()
 	delta := new(big.Int).Sub(amountEvmDenom, balance)
+	bk := k.Bank.BaseKeeper
```
Unique-Divine, line 78: The `Bank.BaseKeeper` needs to be used here because we're inside the setter that the `StateDB` uses to commit balances. Using the `NibiruBankKeeper`'s overrides would double-label and mutate the journal while its being committed, leading to strange behavior.

CHANGELOG.md
Unique-Divine, line 1: Trying to better about having meaningful changelog entry descriptions and keep it organized rather than simply copying PR titles.

x/evm/keeper/bank_extension.go
```diff
@@ -0,0 +1,155 @@
...
+	moduleName string,
+	coins sdk.Coins,
+) error {
+	// Use the embedded function from [bankkeeper.Keeper]
+	if err := bk.BaseKeeper.MintCoins(ctx, moduleName, coins); err != nil {
+		return err
+	}
+	if findEtherBalanceChangeFromCoins(coins) {
+		moduleBech32Addr := auth.NewModuleAddress(evm.ModuleName)
+		bk.SyncStateDBWithAccount(ctx, moduleBech32Addr)
```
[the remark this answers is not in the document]
Unique-Divine: We have to follow the original Bank Keeper's interface exactly, or the override won't work as intended

x/evm/keeper/bank_extension.go
```diff
@@ -0,0 +1,155 @@
...
+	moduleName string,
+	coins sdk.Coins,
+) error {
+	// Use the embedded function from [bankkeeper.Keeper]
+	if err := bk.BaseKeeper.BurnCoins(ctx, moduleName, coins); err != nil {
+		return err
+	}
+	if findEtherBalanceChangeFromCoins(coins) {
+		moduleBech32Addr := auth.NewModuleAddress(evm.ModuleName)
+		bk.SyncStateDBWithAccount(ctx, moduleBech32Addr)
```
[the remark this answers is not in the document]
Unique-Divine: Same reasoning as here https://github.com/NibiruChain/nibiru/pull/2095#discussion_r1824330443

x/evm/keeper/bank_extension.go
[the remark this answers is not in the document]
k-yang: In that case, I think we should use `auth.NewModuleAddress(moduleName)` instead of `auth.NewModuleAddress(evm.ModuleName)` because the `MintCoins()` caller can pass in any module name, and it would be incorrect to always sync the balance changes to the `evm.ModuleName`.

x/evm/keeper/funtoken_from_coin_test.go
k-yang: One improvement could be to add a field to `FunTokenBalanceAssert ` for a custom description message, such as `description: "Alice should have zero NIBI"`. It would make reading through a series of struct assertions much quicker imo.

x/evm/precompile/funtoken.go
```diff
@@ -151,7 +146,12 @@ func (p precompileFunToken) bankSend(
 			return
 		}
 	} else {
-		err = SafeMintCoins(ctx, evm.ModuleName, coinToSend, p.bankKeeper, start.StateDB)
+		// NOTE: The NibiruBankKeeper needs to reference the current [vm.StateDB] before
+		// any operation that has the potential to use Bank send methods. This will
+		// guarantee that [evmkeeper.Keeper.SetAccBalance] journal changes are
+		// recorded if wei (NIBI) is transferred.
+		p.evmKeeper.Bank.StateDB = start.StateDB
```
k-yang, line 153: Didn't we already set `p.evmKeeper.Bank.StateDB` in the `Run()` method, prior to calling `bankSend`? I think it's better to keep it in the `Run()` method because that executes prior to any specific method call.

x/evm/precompile/funtoken.go
```diff
@@ -160,13 +160,17 @@ func (p precompileFunToken) bankSend(
 	}
 
 	// Transfer the bank coin
-	err = SafeSendCoinFromModuleToAccount(
+	//
+	// NOTE: The NibiruBankKeeper needs to reference the current [vm.StateDB] before
+	// any operation that has the potential to use Bank send methods. This will
+	// guarantee that [evmkeeper.Keeper.SetAccBalance] journal changes are
+	// recorded if wei (NIBI) is transferred.
+	p.evmKeeper.Bank.StateDB = start.StateDB
```
k-yang, line 168: Likewise, `p.evmKeeper.Bank.StateDB` should already be set, so there shouldn't be a need to set it again.

x/evm/keeper/bank_extension.go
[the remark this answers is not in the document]
Unique-Divine: You just pass the exact argument to the base keeper. No need to use auth

x/evm/precompile/funtoken.go
Unique-Divine: It gets messed up by the ERC20 tranfers and stuff

x/evm/keeper/bank_extension.go
```diff
@@ -0,0 +1,155 @@
...
+	// Use the embedded function from [bankkeeper.Keeper]
+	if err := bk.BaseKeeper.MintCoins(ctx, moduleName, coins); err != nil {
+		return err
+	}
+	if findEtherBalanceChangeFromCoins(coins) {
+		moduleBech32Addr := auth.NewModuleAddress(evm.ModuleName)
+		bk.SyncStateDBWithAccount(ctx, moduleBech32Addr)
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
Unique-Divine: They have to use non-pointer receivers for the `bankkeeper.Keeper` interface

x/evm/keeper/funtoken_from_coin_test.go
Unique-Divine: Yeah, I like that. And since it's a struct, the field is actually optional. Adding now

k-yang (approved):

x/evm/precompile/funtoken.go
Unique-Divine: It ends up creating a new reference to the *statedb.StateDB entirely. You'll find that if you remove these reassignments, it actually regresses and reintroduces the security vulnerability, failing our unit tests.

This test is the meaningful one:  `TestNativeSendThenPrecompileSend`

This issue arises from both `ERC20().Transfer` and `ERC20().Burn` calling ApplyEvmMsg with commit=true. This results in calling 
```go
	stateDB := k.NewStateDB(ctx, txConfig)
```
and creates a fresh pointer that differs from the parent StateDB that the precompile uses. Sadly, editing the child StateDB's journal doesn't affect the parent. 

And yes, adding the pointer assignment is also inexpensive

PR #2110, opened 2024-11-11
fix(evm): non-deterministic gas usage

k-yang:
# Purpose / Abstract

- #2095 introduced a custom Nibiru BankKeeper extension that adds StateDB as a pointer field. The pointer field is updated in readonly query endpoints (e.g. eth_estimateGas) which makes gas usage non-deterministic in future transactions because of the conditional program flow path arising from a nil check. The conditional check causes consensus failures because of the difference in gas used.
- The solution is to restore the StateDB to what it was prior to the ApplyEvmMsg call, after the ApplyEvmMsg call.
- How I tested it on localnet:
  1) measure gas usage of a simple x/bank send
  2) manually call the `eth_estimateGas` endpoint to trigger a BankKeeper.StateDB pointer update
  3) measure gas usage of x/bank send again
  4) prior to the fix (`v2.0.0-rc.12`), the gas_used is different. On this branch, the gas_used is identical.

See https://www.notion.so/nibiru/2024-11-07-testnet-1-consensus-failure-137d2597a03c8064a399eec8f24614aa?pvs=4 for more details

## Release Notes

- **New Features**
	- Enabled WASM light clients on IBC.
	- Introduced support for foundry in Nibiru EVM development.
	- Added validation for FunToken creation fees.
	- Enhanced functionality for gas usage in precompiles.
	- Implemented a new method for retrieving fun token mappings.

- **Bug Fixes**
	- Improved handling of ERC20 transfers with false success values.
	- Ensured state consistency in precompile execution.
	- Addressed gas consumption issues during ERC20 contract execution.
	- Enhanced error handling and removed deprecated API keys.

- **Documentation**
	- Updated CHANGELOG.md to reflect recent changes and improvements.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -104,7 +104,7 @@ tests for race conditions within funtoken precompile
 - [#2107](https://github.com/NibiruChain/nibiru/pull/2107) -
 feat(evm-funtoken-precompile): Implement methods: balance, bankBalance, whoAmI
 - [#2108](https://github.com/NibiruChain/nibiru/pull/2108) - fix(evm): removed deprecated root key from eth_getTransactionReceipt
-
+- [#2108](https://github.com/NibiruChain/nibiru/pull/2110) - fix(evm): get rid of non-deterministic gas usage of NibiruBankKeeper
 
 #### Nibiru EVM | Before Audit 1 - 2024-10-18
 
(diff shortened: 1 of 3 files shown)


Unique-Divine (approved):
lgtm

flackoon:
Hi @k-yang, I'm coming here from the recent competition on code4rena. I'm trying to understand how does the `eth_estimateGas` resetting the `StateDB` pointer affect the gas usage of the second `x/bank` Send message. I see that `SyncStateDBWithAccount` will update the account balances in the `StateDB` and will generate journal entries, but that's about all I see. I had no success searching for any gas calculation mechanism that would be affected by the `eth_estimateGas`. If you could point it out I'd be more than thankful.

PR #2122, opened 2024-12-11
test(evm): more bank extension tests and EVM ABCI integration tests to prevent regressions

Unique-Divine:
# Purpose / Abstract

- 1. It looked like there was redundant gas subtraction in `EthereumTx`
- 2. To take a closer look at (1), I added a test that uses the raw ABCI methods instead of only using the message server directly. There's now a test that uses raw EthereumTx bytes in `abci.BeginBlock -> abci.DeliverTx -> abci.EndBlock` and displays proper gas consumption. → In doing this, I found out that nothing in our current test suite prevents a regression if we subtract out the consume gas step after an `EthereumTx`.  
- 3. We were using a branched context for `EthereumTx > ApplyEvmMsg` unnecessarily. The `BaseApp` already uses a temporary context and passes it in for `DeliverTx`, so there's not need to worry about cleanup in `ApplyEvmMsg`. 
- 4. Clarifies TODO comments in the precompiles directory
- 5. Further tests have been added for the `NibiruBankKeeper` to verify it's not going to bring about unexpected state changes

### DIFF
--- a/x/evm/keeper/msg_server.go
+++ b/x/evm/keeper/msg_server.go
@@ -59,10 +59,10 @@ func (k *Keeper) EthereumTx(
 		return nil, errors.Wrap(err, "failed to return ethereum transaction as core message")
 	}
 
-	tmpCtx, commitCtx := ctx.CacheContext()
-
-	// pass true to commit the StateDB
-	evmResp, _, err = k.ApplyEvmMsg(tmpCtx, evmMsg, nil, true, evmConfig, txConfig, false)
+	// ApplyEvmMsg - Perform the EVM State transition
+	refundLeftoverGas := false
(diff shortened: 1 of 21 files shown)


x/evm/keeper/msg_server.go
```diff
@@ -269,7 +271,7 @@ func (k *Keeper) ApplyEvmMsg(ctx sdk.Context,
 	)
 
 	defer func() {
-		if commit && err == nil && resp != nil && !resp.Failed() {
```
Unique-Divine, line 272: It's perfectly valid to have `err == nil` while the `vmError` is not empty. That will still result in a valid transaction.

PR #2132, opened 2024-12-26
fix(evm): proper tx gas refund outside of CallContract

onikonychev:
- resolves https://github.com/code-423n4/2024-11-nibiru-findings/issues/45

## Summary by CodeRabbit

- **Bug Fixes**
	- Fixed transaction gas refund mechanism in the EVM component.
	- Improved error handling and gas management in Ethereum transaction processing.
	- Enhanced gas calculation accuracy for contract interactions.

- **Testing**
	- Added new test suites for gas usage and transaction handling.
	- Improved concurrency control in test environments.
	- Updated test methods to handle more comprehensive transaction scenarios.

- **Documentation**
	- Updated CHANGELOG.md with latest bug fix details.


k-yang (approved):

Unique-Divine (changes requested):
Based on the failing test(s), this change seems to introduce an issue where a required refund is bigger than what's given as input by the msg sender. 

The issue may be resolved by broadcasting the message with a full transaction rather than using gRPC keeper methods directly, as that will guarantee the ante handler runs before the new gas logic

x/evm/keeper/msg_server.go
```diff
@@ -563,8 +562,15 @@ func (k Keeper) convertCoinToEvmBornCoin(
...
 	)
 	if err != nil {
+		k.ResetGasMeterAndConsumeGas(ctx, ctx.GasMeter().Limit())
 		return nil, err
 	}
+	blockGasUsed, errBlockGasUsed := k.AddToBlockGasUsed(ctx, evmResp.GasUsed)
+	if errBlockGasUsed != nil {
+		return nil, errors.Wrap(errBlockGasUsed, "error adding transient gas used")
+	}
+	k.ResetGasMeterAndConsumeGas(ctx, blockGasUsed)
```
k-yang, line 572: Shouldn't we consume `evmResp.GasUsed` instead of `blockGasUsed` here?

Unique-Divine (approved):

PR #2139, opened 2025-01-03
fix(evm): erc20 born funtoken: properly burn bank coins after convert…

onikonychev:
- resolves https://github.com/code-423n4/2024-11-nibiru-findings/issues/48

## Summary by CodeRabbit

- **Bug Fixes**
	- Fixed an issue with ERC20 born funtoken to ensure bank coins are properly burned after converting back to ERC20.

- **New Features**
	- Added a new test ERC20 contract with a transfer fee mechanism.
	- Introduced additional test coverage for ERC20 token conversion.

- **Documentation**
	- Updated changelog with details of the bug fix and changes.


k-yang (approved):

x/evm/keeper/msg_server.go
Unique-Divine, line 1: This looks like the re-introduction of a bug fixed in the first audit. Please explain further why the sent amount is being ignored now

Unique-Divine (changes requested):

x/evm/keeper/msg_server.go
k-yang: Here's my take on the issue here: https://github.com/code-423n4/2024-11-nibiru-findings/issues/48#issuecomment-2573453351

matthiasmatt: @Unique-Divine I think we should burn them (approve this PR). For reference, this types of transfer function are usually either:
- Burned (sent to a dead address).
- Reflected (distributed to existing holders).
- Routed to a Marketing/Dev Wallet (for project funding).
- Used for Liquidity (auto-liquidity functions).
- Any combination of the above (split fees).


In any of these cases, all of the tokens are expected to be used by the function and derived to wallets, so we can burn all the amount.

As for the previous fix, it was related to this code but not exactly the same. We use to check that input in = input sent to the transferee wallet, which is not correct considering these weird fee-on-transfer tokens, and that's why we made it more lax and don't check this anymore.

Unique-Divine (approved):

PR #2140, opened 2025-01-03
fix(bank): fix gas invariant wrapper to actually charge gas

k-yang:
The bank keeper extension was charging zero gas for its operations.

Related: #2119 

## Summary by CodeRabbit

- **New Features**
	- Added a method to synchronize state database with account balances.
	- Introduced a function to detect Ether balance changes.

- **Bug Fixes**
	- Enhanced gas consumption logic to prevent panics during operations.
	- Adjusted expected gas consumption in specific test cases.

- **Tests**
	- Updated gas consumption test cases with more precise assertions.
	- Added logging for gas consumption values.
	- Improved clarity of assertions related to gas consumption invariants.

- **Documentation**
	- Updated changelog entries for clarity and organization, including new sections for recent changes.


Unique-Divine:
It is charging gas already. What the function does is only factor in the gas from using the underlying base bank keeper. I don't think you should remove that

Unique-Divine (approved):

Issue #2158, opened 2025-01-10
[Bug] Attacker has the ability to send funds to module address, potentially breaking blockchain invariants and halting the network


stonejiajia:
## Root Cause

A critical vulnerability in the bankMsgSend precompile function allows attackers to send funds directly to module address. This capability can potentially break blockchain invariants and lead to a network halt. The lack of proper checks in the function enables malicious actors to disrupt the expected state of the system, posing a severe threat to the network's stability and integrity.

The vulnerability is triggered through the following execution path:
FunTokenMethod_bankMsgSend -> bankMsgSend -> sendCoins -> sendCoins ( cosmos )

```go
func (bk NibiruBankKeeper) SendCoins(
	ctx sdk.Context,
	fromAddr sdk.AccAddress,
	toAddr sdk.AccAddress,
	coins sdk.Coins,
) error {
	return bk.ForceGasInvariant(
		ctx,
		func(ctx sdk.Context) error {
			return bk.BaseKeeper.SendCoins(ctx, fromAddr, toAddr, coins)
		},
		func(ctx sdk.Context) {
			if findEtherBalanceChangeFromCoins(coins) {
				bk.SyncStateDBWithAccount(ctx, fromAddr)
				bk.SyncStateDBWithAccount(ctx, toAddr)
			}
		},
	)
}
```

The final step uses the Cosmos bank keeper function SendCoins. In Cosmos, there is no validation of the toAddr in SendCoins. Only the send function validates the toAddr

```go
func (k msgServer) Send(ctx context.Context, msg *types.MsgSend) (*types.MsgSendResponse, error) {
        .....
	if k.BlockedAddr(to) {
		return nil, errorsmod.Wrapf(sdkerrors.ErrUnauthorized, "%s is not allowed to receive funds", msg.ToAddress)
	}

	err = k.SendCoins(ctx, from, to, msg.Amount)
	if err != nil {
		return nil, err
	}

	return &types.MsgSendResponse{}, nil
}
```

[Document](https://docs.cosmos.network/v0.46/modules/bank/02_keepers.html#common-types) Explain that:

> The x/bank module accepts a map of addresses that are considered blocklisted from directly and explicitly receiving funds.

> Typically, these addresses are module accounts. If these addresses receive funds outside the expected rules of the state machine, invariants are likely to be broken and could result in a halted network.
### Internal pre-conditions
1. The `bankMsgSend` precompile function needs to be active and accessible in the EVM environment.
[...]

PR #2160, opened 2025-01-10
fix(evm-precompile): use bank.MsgServer Send in precompile IFunToken.bankMsgSend

Unique-Divine:
# Purpose / Abstract

- Fixes https://github.com/NibiruChain/nibiru/issues/2158

The 3 methods where the above ticket might potentially play a role are `sendToBank`, `sendToEvm` and `bankMsgSend`. 

- `sendToBank`  does not need any changes because it uses "SendCoinsFromModuleToAccount" to send Bank Coins, and this method correctly avoids sending funds to addresses blocked by the Bank module.
- Similarly, `sendToEvm` does not need changes because the only bank.MsgSend operation involved is when the signer transfers funds to the EVM module 
- `bankMsgSend` does include a change because `BaseSendKeeper.Send` does not have the blocked address check, so I've switched that to use `bank.MsgServer.Send`, as it's safer and includes the blocking behavior.

### DIFF
--- a/x/evm/precompile/funtoken.go
+++ b/x/evm/precompile/funtoken.go
@@ -6,6 +6,8 @@ import (
 
 	"cosmossdk.io/math"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	bankkeeper "github.com/cosmos/cosmos-sdk/x/bank/keeper"
+	bank "github.com/cosmos/cosmos-sdk/x/bank/types"
 	gethabi "github.com/ethereum/go-ethereum/accounts/abi"
 	gethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/vm"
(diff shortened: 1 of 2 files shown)


x/evm/precompile/funtoken.go
```diff
@@ -715,8 +718,16 @@ func (p precompileFunToken) bankMsgSend(
...
+	bankMsg := &bank.MsgSend{
+		FromAddress: fromBech32.String(),
+		ToAddress:   toBech32.String(),
+		Amount:      coin,
+	}
+	if err := bankMsg.ValidateBasic(); err != nil {
+		return nil, err
+	}
+	if _, err := bankkeeper.NewMsgServerImpl(p.evmKeeper.Bank).Send(
+		sdk.WrapSDKContext(ctx), bankMsg,
```
Unique-Divine, line 730: This adds the block address check to `IFunToken.bankMsgSend`

PR #2163, opened 2025-01-12
feat: reduce gas for first vote by validators

### LINKED ISSUE #1592: feat(oracle): oracle gas needs to be low/negligible
## Problem

Currently, votes for `x/oracle` are expensive for node operators even with lowered fixed gas. Currently, it costs 500 unibi and is dictated by the `OracleMessageGas` constant from `ante/fixed_gas.go`. Although this gas cost is low compared to something like `MsgMarketOrder` (200k unibi), it adds up to be a lot due to the frequency of transactions that have to be sent for the oracle voting rounds. 

Prohibitive costs on `x/oracle` make it unlikely for us to have full participation from every active validator, so we need to make it more cost-efficient without increasing the potential for spam. 

## MVP Solution

Add a module-specific ante handler similar to the one in `x/devgas/ante` that has access to the `OracleKeeper`. When a tx comes in:
1. Check if it has an oracle pre-vote or vote msg. Otherwise, don't change its gas (exit).
2. Check if the sender is an eligible oracle voter (active validator or delegated feeder). Otherwise, exit.
3. Ensure that the tx only has one message. Otherwise, exit.
4. If the oracle has not already voted or prevoted at the current `ctx.BlockHeight` and `voteHash`, lower the gas cost significantly.  

## Open Problems

Spam prevention: We need to make sure that a single active validator can't troll by filling up blocks with low-gas oracle transactions. This becomes more challenging as we lower the gas for the oracle voting messages.

What ways can step 4 of the MVP solution be exploited to spam the network? 

---

Log:
- 2023-09-18: Description follows a recent discussion with @jgimeno

matthiasmatt:
# Purpose / Abstract

- Closes #1592

---

This pull request includes several changes to the `Nibiru` application, focusing on adding a new fee discount mechanism for validators who vote in the current period, updating the `AnteHandler` options, and enhancing test utilities. The most important changes include the addition of the `VoteFeeDiscountDecorator`, updates to the `AnteHandlerOptions` struct, and the implementation of new test utilities.

### Fee Discount Mechanism:
* Added `VoteFeeDiscountDecorator` to apply a fee discount for validators who vote in the current period (`x/oracle/ante/fee_discount.go`).
* Created tests for `VoteFeeDiscountDecorator` to ensure correct functionality (`x/oracle/ante/fee_discount_test.go`).
* Added interfaces for `StakingKeeper` and `OracleKeeper` to be used in the `VoteFeeDiscountDecorator` (`x/oracle/ante/keeper_interface.go`).

### AnteHandler Options:
* Updated `AnteHandlerOptions` struct to include `OracleKeeper` and `StakingKeeper` (`app/ante/handler_opts.go`).
* Modified `NewAnteHandlerNonEVM` to include the new `VoteFeeDiscountDecorator` (`app/ante.go`).

### Test Utilities:
* Added `AddTestAddrsIncremental` function to generate test addresses and optionally fund them (`x/common/testutil/testapp/testapp.go`).

### Additional Changes:
* Added imports for `oracleante` and `stakingkeeper` in relevant files (`app/ante.go`, `app/ante/handler_opts.go`). [[1]](diffhunk://#diff-5c8651424dbfaa86b2894ea499273a1b26c4e0abf0d534b6b9e103b0822cb2b1R12-R13) [[2]](diffhunk://#diff-4aa5ed39605e30ee5b9c59ab9d47dca7007ca1e7aa2858fe8a50bfaf6dce3f94R13-R18)
* Updated `NewNibiruApp` to include `OracleKeeper` and `StakingKeeper` in the options (`app/app.go`).

These changes enhance the functionality and testability of the `Nibiru` application by introducing a new fee discount mechanism for validators and improving the configuration and testing capabilities of the application.


Unique-Divine:
This is great. We've had this ticket open a while

Unique-Divine:
Note to merge after V2

x/oracle/ante/fee_discount.go
```diff
@@ -0,0 +1,102 @@
...
+		if _, ok := msg.(*types.MsgAggregateExchangeRatePrevote); !ok {
+			if _, ok := msg.(*types.MsgAggregateExchangeRateVote); !ok {
+				return next(ctx, tx, simulate)
+			}
+		}
+	}
+
+	// 2. Check if the signer is a validator
+	valAddr := sdk.ValAddress(sigTx.GetSigners()[0])
+	validator, found := vfd.stakingKeeper.GetValidator(ctx, valAddr)
```
k-yang, line 70: This won't work because we let validators delegate their price voting authority to another account (aka the pricefeeder).

x/oracle/ante/fee_discount.go
```diff
@@ -0,0 +1,102 @@
...
+	if !found {
+		return next(ctx, tx, simulate)
+	}
+
+	if validator.Jailed {
+		return next(ctx, tx, simulate)
+	}
+
+	// needs to have at least 0.5% of the supply bonded
+	totalBonded := vfd.stakingKeeper.TotalBondedTokens(ctx)
```
k-yang, line 80: Completely arbitrary, shouldn't we let any validator obtain this gas fee discount, regardless of bonded amount?

x/oracle/ante/fee_discount.go
```diff
@@ -0,0 +1,102 @@
...
+
+	if currentlyBonded.LT(totalBonded.Mul(math.NewInt(50)).Quo(math.NewInt(10000))) {
+		return next(ctx, tx, simulate)
+	}
+
+	// 3. If validator, we check whether they've posted a vote this period
+	hasVoted := vfd.oracleKeeper.HasVotedInCurrentPeriod(ctx, valAddr)
+	if !hasVoted {
+		// 4. If first time, let's say we reduce gas cost.
+		minGasPrices := ctx.MinGasPrices()
```
k-yang, line 91: This won't actually trigger a lower fee for the transaction sender. `ctx.MinGasPrices()` is a validator local config ONLY USED for determining whether or not the tx makes it into the validator's mempool.

Once the block has been proposed and consensus voted on, the txs get delivered to the app (via ABCI DeliverTx()) and the exact fee from the tx is used. Local validator gas prices are not considered in `DeliverTx()`.

k-yang (changes requested):

x/oracle/ante/fee_discount.go
matthiasmatt: I could refund the gas in this case?

x/oracle/ante/fee_discount.go
matthiasmatt: My reasoning is that we would want to avoid someone joining in and spamming the chain with low bonded tokens. I can remove it if needed

PR #2165, opened 2025-01-14
fix(evm): ensure only one copy of StateDB when executing Ethereum txs

k-yang:
# Purpose / Abstract

- Closes https://github.com/code-423n4/2024-11-nibiru-findings/issues/24


app/evmante/evmante_can_transfer.go
```diff
@@ -59,37 +57,27 @@ func (ctd CanTransferDecorator) AnteHandle(
...
-		cfg := &statedb.EVMConfig{
-			ChainConfig: ethCfg,
-			Params:      params,
-			// Note that we use an empty coinbase here  because the field is not
-			// used during this Ante Handler.
-			BlockCoinbase: gethcommon.Address{},
-			BaseFeeWei:    baseFeeWeiPerGas,
-		}
-
-		stateDB := ctd.NewStateDB(
```
k-yang, line 75: Creating a new StateDB here uses the anteCtx, which gets tossed when the tx runs. That means whenever StateDB commits, it's committing to a out of scope ctx.

It's better to just read the bank balances directly via the Cosmos ctx (which is what core.CanTransfer does anyways).

eth/rpc/backend/backend_suite_test.go
```diff
@@ -173,7 +173,7 @@ func SendTransaction(s *BackendSuite, tx *gethcore.LegacyTx, waitForNextBlock bo
 
 // WaitForReceipt waits for a transaction to be included in a block, returns block number, block hash and receipt
 func WaitForReceipt(s *BackendSuite, txHash gethcommon.Hash) (*big.Int, *gethcommon.Hash, *backend.TransactionReceipt) {
-	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
```
k-yang, line 176: I increased the wait time because a 5 second timeout was causing issues when I stepped through the code with a debugger (which takes longer than 5 seconds).

eth/rpc/backend/gas_used_test.go
```diff
@@ -154,10 +154,6 @@ func (s *BackendSuite) TestGasUsedFunTokens() {
 	s.Require().NotNil(blockNumber2)
 	s.Require().NotNil(blockNumber3)
 
-	// TXs should have been included in the same block
-	s.Require().Equal(blockNumber1, blockNumber2)
```
k-yang, line 158: Unnecessary to have the txs in the same block.

x/common/testutil/testnetwork/network.go
```diff
@@ -494,7 +494,7 @@ func (n *Network) LatestHeight() (int64, error) {
 // committed after a given block. If that height is not reached within a timeout,
 // an error is returned. Regardless, the latest height queried is returned.
 func (n *Network) WaitForHeight(h int64) (int64, error) {
-	return n.WaitForHeightWithTimeout(h, 40*time.Second)
+	return n.WaitForHeightWithTimeout(h, 5*time.Minute)
```
k-yang, line 497: Same issue with the debugger timing out.

x/evm/keeper/call_contract.go
```diff
@@ -7,46 +7,13 @@ import (
...
-//   - abi: The ABI (Application Binary Interface) of the smart contract.
-//   - fromAcc: The Ethereum address of the account initiating the contract call.
-//   - contract: Pointer to the Ethereum address of the contract to be called.
-//   - commit: Boolean flag indicating whether to commit the transaction (true) or simulate it (false).
-//   - methodName: The name of the contract method to be called.
-//   - args: Variadic parameter for the arguments to be passed to the contract method.
-//
-// Note: This function handles both contract method calls and simulations,
-// depending on the 'commit' parameter.
-func (k Keeper) CallContract(
```
k-yang, line 32: I realized that `CallContract` is basically `CallContractWithInput`, but `CallContractWithInput` is more generic because it allows for contract deployments. I refactored the code to just use `CallContractWithInput` and removed `CallContract` to reduce the API surface of `EVMKeeper`.

x/evm/keeper/call_contract.go
```diff
@@ -63,12 +30,13 @@ func (k Keeper) CallContract(
 // depending on the 'commit' parameter. It uses a default gas limit.
 func (k Keeper) CallContractWithInput(
 	ctx sdk.Context,
+	evmObj *vm.EVM,
 	fromAcc gethcommon.Address,
 	contract *gethcommon.Address,
 	commit bool,
 	contractInput []byte,
 	gasLimit uint64,
-) (evmResp *evm.MsgEthereumTxResponse, evmObj *vm.EVM, err error) {
```
k-yang, line 71: We don't need to return `evmObj` anymore because it comes from the upper scope.

x/evm/keeper/erc20.go
k-yang, line 1: Basically we pass in `evmObj` now from the upper scope.

x/evm/keeper/funtoken_from_coin.go
```diff
@@ -75,16 +77,43 @@ func (k *Keeper) deployERC20ForBankCoin(
...
 		k.ResetGasMeterAndConsumeGas(ctx, ctx.GasMeter().Limit())
 		return gethcommon.Address{}, errors.Wrap(err, "failed to deploy ERC20 contract")
 	}
+
+	err = stateDB.Commit()
+	if err != nil {
+		return gethcommon.Address{}, errors.Wrap(err, "failed to commit stateDB")
+	}
+	// Don't need the StateDB anymore because it's not usable after committing
+	k.Bank.StateDB = nil
```
k-yang, line 115: We always have to set the `Bank.StateDB` to nil so that the StateDB doesn't accrue pending changes from the BeginBlocker/EndBlocker hooks (e.g. distribution and fee_collector balance changes).

Plus, there's no guarantee that StateDB is usable after `Commit()` is called, so it's better to just let the Golang garbage collector dispose of it.

x/evm/precompile/funtoken.go
```diff
@@ -69,15 +69,15 @@ func (p precompileFunToken) Run(
 	method := startResult.Method
 	switch PrecompileMethod(method.Name) {
 	case FunTokenMethod_sendToBank:
-		bz, err = p.sendToBank(startResult, contract.CallerAddress, readonly)
+		bz, err = p.sendToBank(startResult, contract.CallerAddress, readonly, evm)
```
k-yang, line 72: We pass in the `evm` object to precompile methods that need access to `StateDB`.

Unique-Divine (approved):
Looks great. Didn't see anything concerning on my side

x/evm/precompile/funtoken.go
Unique-Divine: Yes, this looks so much cleaner and less error-prone

x/evm/keeper/funtoken_from_coin_test.go
```diff
@@ -27,7 +27,8 @@ func (s *FunTokenFromCoinSuite) TestCreateFunTokenFromCoin() {
 	// Compute contract address. FindERC20 should fail
 	nonce := deps.NewStateDB().GetNonce(deps.Sender.EthAddr)
 	contractAddress := crypto.CreateAddress(deps.Sender.EthAddr, nonce)
-	metadata, err := deps.EvmKeeper.FindERC20Metadata(deps.Ctx, contractAddress)
+	evmObj, _ := deps.NewEVM()
```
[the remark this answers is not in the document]
k-yang: deps.NewEVM() does not return an error

Unique-Divine:
Almost done resolving conflicts. I'm waiting for the tests to run

PR #2171, opened 2025-01-21
chore: add checktx for AnteDecVerifyEthAcc

expertdicer:
# Purpose / Abstract

Based on [QA-11 Missing CheckTx optimization in account verification leads to redundant processing](https://github.com/code-423n4/2024-11-nibiru-findings/blob/main/data/Bauchibred-Q.md#qa-11-missing-checktx-optimization-in-account-verification-leads-to-redundant-processing)


Unique-Divine:
What is the URL for the original issue?

k-yang (changes requested):
I actually don't think this is a good idea to add in (the QA report and Ethermint did not think through this critically enough).

`CheckTx` is only called when a tx arrives at the local mempool, but not at block execution time. Only `DeliverTx` is called at block execution time. 

That means a malicious validator could include a tx in the proposed block that doesn't pass this AnteHandler check, and we'd bypass through that check if we merge this PR.

Unique-Divine:
Agreed on this point from @k-yang (https://github.com/NibiruChain/nibiru/pull/2171#pullrequestreview-2565090891). I think we should not include this change

PR #2180, opened 2025-01-27
fix(evm): consume gas in CallContractWithInput

k-yang:
# Purpose / Abstract

- Closes https://github.com/zenith-security/2025-01-nibiru-mr-zenith/issues/3
- Closes https://github.com/zenith-security/2025-01-nibiru-mr-zenith/issues/4

Ensures consistent gas consumption in all areas of the EVM codebase by applying gas consumption at the `CallContractWithInput` layer.


x/evm/keeper/call_contract.go
```diff
@@ -39,7 +39,7 @@ func (k Keeper) CallContractWithInput(
 ) (evmResp *evm.MsgEthereumTxResponse, err error) {
 	// This is a `defer` pattern to add behavior that runs in the case that the
 	// error is non-nil, creating a concise way to add extra information.
-	defer HandleOutOfGasPanic(&err, "CallContractError")
+	defer HandleOutOfGasPanic(&err, "CallContractError")()
```
k-yang, line 42: missing parentheses was causing issues with catching OOG errors lol

x/evm/keeper/call_contract.go
```diff
@@ -63,9 +63,11 @@ func (k Keeper) CallContractWithInput(
 	evmResp, err = k.ApplyEvmMsg(
 		ctx, evmMsg, evmObj, evm.NewNoOpTracer(), commit, txConfig.TxHash, true,
 	)
+	if evmResp != nil {
+		ctx.GasMeter().ConsumeGas(evmResp.GasUsed, "CallContractWithInput")
```
k-yang, line 67: We should always consume gas after `ApplyEvmMsg()`, before returning away from `CallContractWithInput()`, or else we lose the opportunity to apply the `EvmResp.GasUsed` variable.

x/evm/keeper/erc20.go
```diff
@@ -196,7 +196,7 @@ func (e erc20Calls) loadERC20String(
 	if err != nil {
 		return out, err
 	}
-	res, err := e.Keeper.CallContractWithInput(
+	evmResp, err := e.Keeper.CallContractWithInput(
```
k-yang, line 199: just a minor rename

x/evm/keeper/funtoken_from_coin.go
```diff
@@ -114,7 +114,5 @@ func (k *Keeper) deployERC20ForBankCoin(
 		return gethcommon.Address{}, errors.Wrap(err, "failed to commit stateDB")
 	}
 
-	ctx.GasMeter().ConsumeGas(evmResp.GasUsed, "deploy erc20 funtoken contract")
```
k-yang, line 117: We no longer need to consume gas here since we consume it at a lower level in `CallContractWithInput()`.

x/evm/keeper/gas_fees.go
```diff
@@ -75,15 +75,29 @@ func (k *Keeper) RefundGas(
...
+// leftoverGas = amount of extra (not used) gas.
+// If the msg comes from user, we apply refundQuotient capping the refund to 20% of used gas
+// If msg is internal (funtoken), we refund 100%
+//
+// EIP-3529: refunds are capped to gasUsed / 5
+// We evaluate "fullRefundLeftoverGas" and use only the gas consumed (not the
+// gas limit) if the `ApplyEvmMsg` call originated from a state transition
+// where the chain set the gas limit and not an end-user.
+func GasToRefund(availableRefundAmount, gasUsed uint64, fullRefundLeftoverGas bool) uint64 {
+	refundQuotient := params.RefundQuotientEIP3529
```
k-yang, line 91: Moved all the 1/5 calculations here.

x/evm/keeper/keeper.go
```diff
@@ -128,13 +128,12 @@ func HandleOutOfGasPanic(err *error, format string) func() {
...
 			switch r.(type) {
 			case sdk.ErrorOutOfGas:
-
 				*err = vm.ErrOutOfGas
 			default:
 				panic(r)
 			}
 		}
-		if err != nil && format != "" {
+		if *err != nil && format != "" {
```
k-yang, line 136: Need to check `*err != nil`, not `err != nil` because `err` is a pointer value.

x/evm/keeper/msg_server.go
```diff
@@ -318,86 +318,62 @@ func (k *Keeper) ApplyEvmMsg(
...
 
-	// Min gas used is a % of gasLimit
-	gasUsed := math.LegacyNewDec(int64(temporaryGasUsed)).TruncateInt().Uint64()
-
-	// This resulting "leftoverGas" is used by the tracer. This happens as a
-	// result of the defer statement near the beginning of the function with
-	// "vm.Tracer".
-	leftoverGas = msg.Gas() - gasUsed
+	// The dirty states in `StateDB` is either committed or discarded after return
+	if commit {
```
k-yang, line 370: I moved the StateDB.Commit() to after the gas calculations so we have accurate gas accounting prior to commit and return.

PR #2190, opened 2025-02-03
fix(evm): handling empty transfer response

expertdicer:
# Purpose / Abstract

- Closes https://github.com/code-423n4/2024-11-nibiru-findings/issues/54


Unique-Divine:
I don't think we should or need to include this change for this reason 

https://github.com/code-423n4/2024-11-nibiru-findings/issues/54#issuecomment-2556574950

Unique-Divine (changes requested):
For example, USDT on newer networks like Base and Kava is deployed as a proper ERC20.

- [USDT on Kavascan](https://kavascan.com/address/0xee3bbc3780a6b4081c52a52ae3eedcdc178a4a75?t=code)
- [USDT on BaseScan](https://basescan.org/token/0xfde4C96c8593536E31F229EA8f37b2ADa2699bb2#code)

PR #2192, opened 2025-02-03
fix(oracle): correctly handle misscount

expertdicer:
# Purpose / Abstract

This pull request includes several changes to the way oracle counting validators who is voting outside the spread.  

Currently, during tallying, if a vote is considered a `miss`, the `missCount` counter increments for each vote. However, it should only increment once per iteration over all votes.

**Solution**: 
Introduce `missedValidators` as a map[string]bool to track whether a validator has already been counted as a miss.

Also added a test `TestTallyMissCount` to confirm the solution works.


x/oracle/keeper/ballot.go
Unique-Divine, line 1: This looks like a correct tally, but why isn't the `missedValidators` map being used anywhere in the oracle logic? It seems like the intention is to refer to that map to fix something else?

Unique-Divine (approved):

PR #2199, opened 2025-02-06
refactor(devgas): For #1866

### LINKED ISSUE #1866: refactor: Use the module name rather than naming all of the exported packages ".../types"
## Context

In older Cosmos-SDK versions, it was common to import the publicly exposed types
from a module by adding a "types" suffix to the package name. For example, 

```go
import (
  banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
  devgastypes "github.com/NibiruChain/nibiru/x/devgas/v1/types"
)
```

## Task List

<details>
<summary>[Expand Task Menu]</summary>

- [ ] refactor(devgas): For #
- [ ] refactor(evm): For #
- [ ] refactor(epochs): For #
- [ ] refactor(genmsg): For #
- [ ] refactor(inflation): For #
- [ ] refactor(oracle): For #
- [ ] refactor(sudo): For #
- [ ] refactor(tokenfactory): For #

</details>

## Implications

1. This means that the true package name, "types", has to be overwritten with an
alias for each import. The Golang LSP (what your IDE understands) still considers
all of the packages as having the name "types", so your IDE will only suggest
variables if you start typing "types". You can't just type "banktypes" in a new
file and see code suggestions.

2. When you use types from two modules at the same time or read code from within
a module, you'll see a nonsemantic prefix, "types" everywhere. Essentially,
package namespacing becomes irrelevant with this approach.

## Suggestion

Follow the pattern of the `nibiru/eth` and `nibiru/x/evm` packages (also the
convention from v0.50 of the Cosmos-SDK) by using the top-level package of a
module for its types.

In other words, use `bank.MsgSend` instead of `types.MsgSend` or
`banktypes.MsgSend`. 

### Support Points:

1. **Less namespace collisions**: The module namespace is often completely unused aside from
one place, which is the top-level "app" directory.

2. **IDEs become more useful**: You can type a module's name and have your IDE automatically suggest its types

3. **Less aliasing**: Using an alias for every import produces a bad code smell (opinionated take).
  > "Using a generic name makes it easy to accidentally use the wrong value in
  > a different file" - [[Uber Golang Style Guide](https://github.com/uber-go/guide/blob/master/style.md)]

  I'd argue that this argument holds for imports and not just for other
  variables.

expertdicer:
# Purpose / Abstract

- Partialy closes #1866


Unique-Divine (changes requested):
The issue #1866 is less about renaming imports and more so suggesting to change the actual package names at the top of the Go files. 

Example: EVM Module
- The module.go and genesis.go are inside `evm/evmmodule` with package name `evmmodule`
- The "types" inside of `evm` with package name `evm`. 

Contrast this with the Devgas module:
- The module.go and genesis.go are inside `devgas/v1` with package name `devgas`. 
- The types are inside of `devgas/v1/types` with package name `types`

To make the imports similar to the EVM one, we will need to update the `.proto` files in devgas to export to the newly desired directory and change the `package types` statements to `package devgas`.

x/devgas/tx.pb.go
```diff
@@ -445,43 +445,43 @@ func init() {
 func init() { proto.RegisterFile("nibiru/devgas/v1/tx.proto", fileDescriptor_72949c99a02cd615) }
 
 var fileDescriptor_72949c99a02cd615 = []byte{
-	// 566 bytes of a gzipped FileDescriptorProto
```
k-yang, line 448: I wonder if this will break the existing state.

k-yang (commented):
The golang package changes look fine to me, but I wonder if the proto package changes will change the byte representation of serialized protos.

Can you test this change by running a localnet with latest `main`, and then stopping the network, hot-swapping to a binary produced by this branch, and starting the network again? I'd like to see if there are any state issues.

Unique-Divine:
Stale, closing as not needed

PR #2220, opened 2025-02-25
fix(evm): improved marshalling of the eth tx receipt

onikonychev:
- added `effectiveGasPrice` setting for all eth txs
- added type

Fixes foundry(cast) tx deployment / execution:
```
cast publish $TX --rpc-url $RPC_URL | jq
```


k-yang:
Related to https://github.com/foundry-rs/foundry/issues/7640

k-yang (approved):

Issue #2235, opened 2025-03-16
feat(evm): failed wasm precompile transaction don't propagate error message


matthiasmatt:
I had an error when creating a transaction for the wasm pre-compile (which came to a contract error in wasm). When querying the evm transaction, i just get failed true with no logs attached.

```
❯ curl -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"debug_traceTransaction","params":["0xd7371d317b2b62591225ecbb852e6a9206a60078f946cdc4311ab04d54f99b10"],"id":1}' http://localhost:8545
{"jsonrpc":"2.0","id":1,"result":{"failed":true,"gas":30000000,"returnValue":"","structLogs":[]}}
```

The only way i was able to see the error was with this debug print:

```golang
	data, err := p.Wasm.Execute(ctx, wasmContract, nibiruCaller, msgArgsBz, funds)
	if err != nil {
		fmt.Printf("Wasm execution failed: %v\n", err)
		return
	}
```

matthiasmatt:
Here's the erorr message from the receipt (wasmcaller)

```
[...]

PR #2249, opened 2025-03-31
fix(evm): resetting gas meter for afterOp in bank extension

onikonychev:
- related to #2235

Resetting gas meter makes `AfterOp` execution of `ForceGasInvariant()` gasless and prevents out of gas panics in txs.


x/evm/keeper/bank_extension.go
```diff
@@ -206,6 +206,7 @@ func (bk NibiruBankKeeper) ForceGasInvariant(
 		return err
 	}
 
+	ctx = ctx.WithGasMeter(sdk.NewInfiniteGasMeter())
```
k-yang, line 209: Should we reset ctx's gas meter to the original `gasMeterBefore` after `AfterOp`?

k-yang: Realized that ctx is passed by value, so any updates to it in this scope won't affect the ctx from the outer scope.

k-yang (approved):

PR #2264, opened 2025-04-09
fix: raise erc20 gas limit deploy

expertdicer:
# Purpose / Abstract

- Raise the gas limit for erc20 deploy to 3_000_000


k-yang:
Already addressed in #2268

PR #2266, opened 2025-04-09
feat: app wiring for evm

expertdicer:
# Purpose / Abstract

- App wiring for `evm` module


app/app.go
```diff
@@ -343,20 +344,12 @@ func NewNibiruApp(
 		wasmtypes.StoreKey,
 		devgastypes.StoreKey,
 		tokenfactorytypes.StoreKey,
-
-		evm.StoreKey,
 	)
-	app.tkeys = sdk.NewTransientStoreKeys(evm.TransientKey)
```
k-yang, line 349: Could also remove the `tkeys` field on `App` entirely.

k-yang (approved):

PR #2267, opened 2025-04-09
feat: update geth to allow precompile delegatecall

matthiasmatt:
The current version of geth has an issue around precompile calls through delegatecall. It forces such calls to be read only before reaching precompiles functions, which makes it impossible to pass the readonly condition.

I update geth to this new tag: https://github.com/NibiruChain/go-ethereum/compare/v1.10.27-nibiru...v1.10.27-nibiru.1


k-yang:
Did you research any specific reason why `delegatecall` was disabled for all preocmpiles in Geth? Seems strange that they would do that... unless there were some deeper reason.

k-yang (approved):

PR #2287, opened 2025-04-20
chore: make function comment match function name

dropbigfish:
# Purpose / Abstract

make function comment match function name

- Closes #AAA


k-yang:
Closing since it seems this was already addressed (no files changed in this PR).

PR #2317, opened 2025-05-30
chore(evm): add MIM & USDC.arb to official erc20 token registry

CalicoNino:
# Purpose / Abstract

- Adding MIM tokens to the ERC20 Token Registry


Unique-Divine (approved):
LGTM. The only remaining step is to add a changelog entry to the file inside of the token registry folder. @CalicoNino

PR #2324, opened 2025-06-03
fix: use the first 20 bytes when decoding the EthAddr

k-yang:
# Purpose / Abstract

- Closes #AAA

### DIFF
--- a/eth/state_encoder.go
+++ b/eth/state_encoder.go
@@ -59,7 +59,7 @@ type keEthAddr struct{}
 
 func (_ keEthAddr) Encode(value gethcommon.Address) []byte { return value.Bytes() }
 func (_ keEthAddr) Decode(bz []byte) (int, gethcommon.Address) {
-	return gethcommon.AddressLength, gethcommon.BytesToAddress(bz)
+	return gethcommon.AddressLength, gethcommon.BytesToAddress(bz[:gethcommon.AddressLength])
 }
 func (_ keEthAddr) Stringify(value gethcommon.Address) string { return value.Hex() }
 


Unique-Divine (commented):
Mind adding a small unit test to show it's working properly? 

It could be a block in the v2.5 upgrade test if that makes sense or be  its own small test

x/evm/keeper/evm_state.go
```diff
@@ -35,7 +34,7 @@ type EvmState struct {
 	//    a piece of state in that contract forms the primary key.
 	//  - value (V): State value bytes.
 	AccState collections.Map[
-		AccStatePrimaryKey, // account (EthAddr) + state key (EthHash)
+		collections.Pair[gethcommon.Address, gethcommon.Hash], // account (EthAddr) + state key (EthHash)
```
k-yang, line 37: I used the raw `collections.Pair` type because it makes it more apparent that it's from the collections struct, rather than something else.

eth/state_encoder.go
```diff
@@ -59,7 +59,7 @@ type keEthAddr struct{}
 
 func (_ keEthAddr) Encode(value gethcommon.Address) []byte { return value.Bytes() }
 func (_ keEthAddr) Decode(bz []byte) (int, gethcommon.Address) {
-	return gethcommon.AddressLength, gethcommon.BytesToAddress(bz)
+	return gethcommon.AddressLength, gethcommon.BytesToAddress(bz[:gethcommon.AddressLength])
```
k-yang, line 62: This was the bug. `bz` was coming in as the entire PairKey (address + storage slot key), and `gethcommon.BytesToAddress` was only taking the right-most 20 bytes, whereas it should be the left-most 20 bytes.

app/upgrades/v2_5_0/v2_5_0.go
```diff
@@ -53,145 +51,44 @@ var Upgrade = upgrades.Upgrade{
 	StoreUpgrades: storetypes.StoreUpgrades{},
 }
 
-// Set of addresses that held stNIBI in ERC20 form prior to the v2.5.0 upgrade
-func MAINNET_STNIBI_HOLDERS() []gethcommon.Address {
```
k-yang, line 57: In my strategy, we actually don't even need to keep track of the original stNIBI holders. It comes from the clever realization that the old and new ERC20 contracts have the exact same storage slot layouts, so as long as we don't mess with the balances mapping in the old ERC20 contract, everybody's balances should automatically be maintained. No need to track everybody's balances in memory.

app/upgrades/v2_5_0/v2_5_0.go
```diff
@@ -53,145 +51,44 @@ var Upgrade = upgrades.Upgrade{
...
-	return holders
-}
-
-var (
-	// MAINNET_STNIBI_ADDR is the (real) hex address of stNIBI on mainnet.
-	MAINNET_STNIBI_ADDR = gethcommon.HexToAddress("0xcA0a9Fb5FBF692fa12fD13c0A900EC56Bb3f0a7b")
-
-	// MAINNET_NIBIRU_SAFE_ADDR: Address of a Gnosis Safe managed by the Nibiru team
-	// See https://nibiscan.io/address/0x22CBd7CbF3b33681abB3Ced4D64d71acB9a9dCd2/contract/6900/code
-	MAINNET_NIBIRU_SAFE_ADDR = gethcommon.HexToAddress("0x22CBd7CbF3b33681abB3Ced4D64d71acB9a9dCd2")
```
k-yang, line 78: Likewise, we no longer need to keep track of excess balances and send them to the Nibiru Gnosis Safe because everybody's balances will be maintained after the upgrade.

app/upgrades/v2_5_0/v2_5_0.go
```diff
@@ -53,145 +51,44 @@ var Upgrade = upgrades.Upgrade{
...
+var MAINNET_STNIBI_ADDR = gethcommon.HexToAddress("0xcA0a9Fb5FBF692fa12fD13c0A900EC56Bb3f0a7b")
 
 func UpgradeStNibiContractOnMainnet(
-	nibiru *keepers.PublicKeepers,
+	keepers *keepers.PublicKeepers,
 	ctx sdk.Context,
 	// erc20Addr is the hex address of stNIBI on mainnet
 	// The upgrade handler takes the address as an argument for testing purposes
-	erc20Addr gethcommon.Address,
+	originalErc20Addr gethcommon.Address,
```
k-yang, line 62: Renamed to distinguish from the new erc20 contract address.

app/upgrades/v2_5_0/v2_5_0.go
```diff
@@ -53,145 +51,44 @@ var Upgrade = upgrades.Upgrade{
...
 
 	// Early return if the address is not a contract
-	accOfContract := nibiru.EvmKeeper.GetAccount(ctx, erc20Addr)
-	if accOfContract == nil || !accOfContract.IsContract() {
+	originalErc20Account := keepers.EvmKeeper.GetAccount(ctx, originalErc20Addr)
+	if originalErc20Account == nil || !originalErc20Account.IsContract() {
 		return nil
 	}
 
-	// -------------------------------------------------------------------------
```
k-yang, line 105: As explained above, we no longer need to query the original token balances anymore.

app/upgrades/v2_5_0/v2_5_0.go
```diff
@@ -53,145 +51,44 @@ var Upgrade = upgrades.Upgrade{
...
-
-		iter := nibiru.EvmKeeper.EvmState.AccState.Iterate(ctx, erc20StateRange)
-		defer iter.Close()
-		for ; iter.Valid(); iter.Next() {
-			key := iter.Key()
-			// Deleting through this iterator is safe because the underlying
-			// `sdk.KVStore.Iterator` uses a snapshot. This means deleting the
-			// current key does not affect the iterator unless you call `Set` on that
-			// key again.
-			_ = nibiru.EvmKeeper.EvmState.AccState.Delete(ctx, key)
```
k-yang, line 185: We no longer need to delete the old erc20 state anymore, as we're already overwriting only the metadata fields later on.

app/upgrades/v2_5_0/v2_5_0.go
```diff
@@ -248,138 +144,110 @@ func UpgradeStNibiContractOnMainnet(
...
 		&evm.EventContractDeployed{
 			Sender:       evmMsg.From.Hex(),
-			ContractAddr: erc20AddrForNewDeploment.Hex(),
+			ContractAddr: newErc20Addr.Hex(),
 		},
 	)
+	fmt.Println("deployed ERC20 contract at", newErc20Addr.Hex())
 
 	// -------------------------------------------------------------------------
-	// STEP 5: Sanity check the new contract with address "erc20AddrForNewDeploment"
```
k-yang, line 286: We don't need to check the metadata on the new ERC20 contract because we only care about it on the original contract, so I moved this check to the end of the upgrade handler.

app/upgrades/v2_5_0/v2_5_0.go
k-yang, line 1: I think it's fine to keep fmt.Println statements in the upgrade handler because it'll only run once, ever, and it could help us debug what went wrong if the upgrade does get botched somehow.

app/upgrades/v2_5_0/v2_5_0.go
```diff
@@ -248,138 +144,110 @@ func UpgradeStNibiContractOnMainnet(
...
-				erc20Addr,              /*erc20Contract*/
-				evm.EVM_MODULE_ADDRESS, /*from*/
-				to,                     /*to*/
-				amount,
-				ctx,
-				evmObj,
-			)
-			if err != nil {
-				return fmt.Errorf("mint erc20 error: %w", err)
+		gotName, _ := keepers.EvmKeeper.ERC20().LoadERC20Name(
```
k-yang, line 223: Here, we check that the original contract's ERC20 metadata is updated.

app/upgrades/v2_5_0/v2_5_0_test.go
```diff
@@ -42,6 +43,20 @@ func (s *Suite) TestUpgrade() {
 		funtoken = evmtest.CreateFunTokenForBankCoin(
 			deps, originalbankMetadata.Base, &s.Suite,
 		)
+
+		// ten holders for testing
+		holders = []gethcommon.Address{
```
k-yang, line 48: Because we don't have a MAINNET_STNIBI_HOLDERS() set anymore, we create one just for testing.

app/upgrades/v2_5_0/v2_5_0_test.go
```diff
@@ -75,30 +87,15 @@ func (s *Suite) TestUpgrade() {
...
-		)
-		s.Require().NoError(err)
-
 		evmObj, _ := deps.NewEVMLessVerboseLogger()
-		totalSupplyErc20, err := deps.EvmKeeper.ERC20().TotalSupply(
-			funtoken.Erc20Addr.Address, deps.Ctx, evmObj)
+		totalSupplyErc20, err := deps.EvmKeeper.ERC20().TotalSupply(funtoken.Erc20Addr.Address, deps.Ctx, evmObj)
 		s.Require().NoError(err)
-		s.Require().Equal("69420", totalSupplyErc20.String())
+		s.Require().Equal("1100", totalSupplyErc20.String())
```
k-yang, line 98: arithmetic series

(1+10)/2 * 20

app/upgrades/v2_5_0/v2_5_0_test.go
```diff
@@ -159,21 +155,20 @@ func (s *Suite) TestUpgrade() {
...
+		s.Equal(2, len(newBankMetadata.DenomUnits))
+		s.Equal("Liquid Staked NIBI", newBankMetadata.Name)
+		s.Equal("stNIBI", newBankMetadata.Symbol)
+		s.Equal(uint32(6), newBankMetadata.DenomUnits[1].Exponent)
 	})
 
 	s.Run("New ERC20 impl: owner should be the EVM module", func() {
-		methodName := "owner"
-		input, err := compiledContract.ABI.Pack(methodName)
+		input, err := compiledContract.ABI.Pack("owner")
```
k-yang, line 171: My Cursor IDE inlined and changed the formatting on some lines automatically lol

app/upgrades/v2_5_0/v2_5_0_test.go
```diff
@@ -188,56 +183,22 @@ func (s *Suite) TestUpgrade() {
...
 		evmObj, _ := deps.NewEVMLessVerboseLogger()
-		for holderAddr, wantBal := range mainnetHolderBals {
-			balErc20, err := deps.EvmKeeper.ERC20().BalanceOf(
-				funtoken.Erc20Addr.Address, holderAddr, deps.Ctx, evmObj)
+		for idx, holderAddr := range holders {
+			balErc20, err := deps.EvmKeeper.ERC20().BalanceOf(funtoken.Erc20Addr.Address, holderAddr, deps.Ctx, evmObj)
 			s.Require().NoError(err)
-			s.Require().Equalf(wantBal.String(), balErc20.String(),
-				"holderAddr %s", holderAddr)
+			s.Require().Equal(strconv.Itoa(20*(idx+1)), balErc20.String())
```
k-yang, line 199: Here's the most important check. It validates that every holder's original balance is still in tact after the upgrade.

Unique-Divine (approved):

PR #2328, opened 2025-06-06
fix(evm): ensure stateDB doesn't persist after upgrade handler

k-yang:
# Purpose / Abstract

- Attempt at fixing the consensus issue we saw on some nodes on our testnet-2 environment

This pull request modifies the `UpgradeStNibiContractOnMainnet` function in `v2_5_0.go` to improve the handling of the `StateDB` object by introducing a conditional check and cleanup mechanism.

Key changes:

### StateDB Handling Improvements:
* Replaced direct instantiation of `StateDB` with a conditional check to use `keepers.EvmKeeper.Bank.StateDB` if available. If not, it falls back to creating a new `StateDB` instance.
* Added a `defer` statement to reset `keepers.EvmKeeper.Bank.StateDB` to `nil` after its usage, ensuring proper cleanup and avoiding potential side effects.

### DIFF
--- a/x/evm/evmtest/test_deps.go
+++ b/x/evm/evmtest/test_deps.go
@@ -48,7 +48,13 @@ func (deps TestDeps) NewStateDB() *statedb.StateDB {
 }
 
 func (deps TestDeps) NewEVM() (*vm.EVM, *statedb.StateDB) {
-	stateDB := deps.EvmKeeper.NewStateDB(deps.Ctx, statedb.NewEmptyTxConfig(gethcommon.BytesToHash(deps.Ctx.HeaderHash())))
+	if deps.EvmKeeper.Bank.StateDB == nil {
+		deps.EvmKeeper.Bank.StateDB = deps.NewStateDB()
+	}
+	stateDB := deps.EvmKeeper.Bank.StateDB
+	defer func() {
+		deps.EvmKeeper.Bank.StateDB = nil
+	}()
 	evmObj := deps.EvmKeeper.NewEVM(
 		deps.Ctx,
 		MOCK_GETH_MESSAGE,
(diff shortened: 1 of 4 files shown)


x/evm/evmtest/test_deps.go
```diff
@@ -48,7 +48,13 @@ func (deps TestDeps) NewStateDB() *statedb.StateDB {
...
 
 func (deps TestDeps) NewEVM() (*vm.EVM, *statedb.StateDB) {
-	stateDB := deps.EvmKeeper.NewStateDB(deps.Ctx, statedb.NewEmptyTxConfig(gethcommon.BytesToHash(deps.Ctx.HeaderHash())))
+	if deps.EvmKeeper.Bank.StateDB == nil {
+		deps.EvmKeeper.Bank.StateDB = deps.NewStateDB()
+	}
+	stateDB := deps.EvmKeeper.Bank.StateDB
+	defer func() {
+		deps.EvmKeeper.Bank.StateDB = nil
+	}()
```
Unique-Divine: You're not necessarily supposed to reset on calls of `NewEVM`, only when you're starting a new `EthereumTx`. In the upgrade handler, it's essentially a transaction, so we need to reset at the end, but not throughout the rest of the codebase uniformly.

I believe these lines are what's breaking the "precompile attack" tests

Unique-Divine (changes requested):

x/evm/evmtest/test_deps.go
k-yang: Yep, was debugging this locally and realized that a lot of tests use NewEVM() multiple times, so clearing the StateDB has some issues. I will mark this PR as draft for now and figure out a more elegant solution.

PR #2340, opened 2025-07-02
fix: evm indexer proper parsing of the start block

onikonychev:


Unique-Divine (approved):
Hard to tell what's happening here. What is this fixing? 

Never mind, I see. It was using maxBlockNumber instead of minBlockNumber by grabbing args[1] instead of args[0].

PR #2351, opened 2025-07-28
fix(evm): isolate statedb sync for simulation in nibiru bank extension

onikonychev:
## StateDB Simulation Isolation Fix

**Problem:** Consensus failures on full nodes caused by EVM simulations (`eth_call`, `eth_estimateGas`) contaminating the shared StateDB, leading to inconsistent account states during real transaction execution.

**Solution:** 
- Mark simulation contexts in query methods
- Use StateDB snapshots to isolate simulation changes in `ForceGasInvariant`
- Revert simulation StateDB modifications while preserving accuracy

**Expected result:** Eliminates consensus failures by preventing simulation contamination while maintaining full query functionality.

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -68,7 +68,7 @@ jobs:
 
       - name: golangci-lint
         if: steps.check_nibiru_go.outputs.nibiru-go == 'true'
-        uses: golangci/golangci-lint-action@v6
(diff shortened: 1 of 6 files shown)


Unique-Divine (changes requested):
Looks great, the only thing left is to include a unit test that populates a `bk.stateDB` and then run simulated EVM txs to show that `bk.StateDB` is unaffected

PR #2353, opened 2025-07-30
refactor(oracle): remove dead code from asset registry 

onikonychev:
- Related https://github.com/NibiruChain/pricefeeder/pull/64


x/common/denoms/denoms.go
```diff
@@ -5,6 +5,7 @@ const ( // stablecoins
 	NUSD = "unusd"
 	USD  = "uusd"
 	USDT = "uusdt"
+	USDA = "uusda"
```
Unique-Divine: Have this written without a "u" prefix to break the curse in the price feeder. Aiming for "usda:usd" 
```suggestion
	USDA = "usda"
```

x/oracle/types/params.go
```diff
@@ -44,6 +44,7 @@ var (
 		asset.Registry.Pair(denoms.ATOM, denoms.USD),
 		asset.Registry.Pair(denoms.USDC, denoms.USD),
 		asset.Registry.Pair(denoms.USDT, denoms.USD),
+		asset.Registry.Pair(denoms.USDA, denoms.USD),
```
k-yang: usda:uusd is a pair that compares `usda` with `uusd`, which has a different order of magnitude. I would recommend doing uusda:uusd to keep it in micro-units.

PR #2357, opened 2025-07-31
fix: proper statedb isolation in nibiru bank_extension

onikonychev:
## Background
Mainnet full nodes periodically fail with errors like this:

> failed to execute message; message index: 0: EthereumTx error: error applying ethereum core message: ApplyEvmMsg: failed to commit stateDB: StateDB error: failed to set account: nibi1fl48vsnmsdzcv85q5d2q4z5ajdha8yu3z7dksy is not allowed to receive funds: unauthorized [cosmos/cosmos-sdk@v0.47.11/x/bank/keeper/keeper.go:325] [NibiruChain/nibiru/v2/x/evm/keeper/msg_server.go:427]

this is from block 26637866

Although Txs in that block don't even have txs which do this.

## What is StateDB again?

This is happening because our `NibiruBankKeeper` (one keeper per app) has a StateDB pointer which is supposed to be set in the beginning of each Ethereum TX and unset right after the TX.

StateDB is a temporary storage of the EVM state modifications. Committing StateDB updates the chain state.

```go
type NibiruBankKeeper struct {
	bankkeeper.BaseKeeper
	StateDB *statedb.StateDB
}
```
The idea behind this (why we are not creating the new statedb in each tx) is the syncronization of the EVM Statedb state with the native accounts state. Our precompiles like FunToken could modify the native chain state like account balances (example `bankMsgSend` function). This is happening outside of the statedb changes and committing the statedb after native changes overrides the account state and leads to a wrong final state.

## How Simulations Corrupt the `NibiruBankKeeper` StateDB

`NibiruBankKeeper.StateDB` is supposed to be only used by TXs and nothing else. TXs are executed sequentially, one after another and could not corrupt the shared statedb because each TX is setting it to null after completion.

But Simulations (EstimateGas, EthCall, TraceTx) run in parallel with TX in multi-thread mode and also use `NibiruBankKeeper`. This means, if a TX is currently running (`NibiruBankKeeper.StateDB != nil`) and some simulation is executed in parallel, it could also modify `StateDB` within `NibiruBankKeeper.SyncStateDBWithAccount()` function. As a result, when TX commits it has a wrong `StateDB` state corrupted by the simulation.

## The Fix

Very simple: we just tag simulation execution explicitly in context:
```go
func (k Keeper) EstimateGasForEvmCallType(
	goCtx context.Context, req *evm.EthCallRequest, fromType evm.CallType,
) (*evm.EstimateGasResponse, error) {
	ctx := sdk.UnwrapSDKContext(goCtx)
	ctx = ctx.WithValue(SimulationContextKey, true)
       ...
}
``` 

using some utility methods
```go
// IsSimulation checks if the context is a simulation context.
func IsSimulation(ctx sdk.Context) bool {
	if val := ctx.Value(SimulationContextKey); val != nil {
		if simulation, ok := val.(bool); ok && simulation {
			return true
		}
	}
	return false
}

// IsDeliverTx checks if we're in DeliverTx, NOT in CheckTx, ReCheckTx, or simulation
func IsDeliverTx(ctx sdk.Context) bool {
	return !ctx.IsCheckTx() && !ctx.IsReCheckTx() && !IsSimulation(ctx)
}
```

And within `SyncStateDBWithAccount` we return quickly if the execution is not `DeliverTx`

```
func (bk *NibiruBankKeeper) SyncStateDBWithAccount(
	ctx sdk.Context, acc sdk.AccAddress,
) {
	// If there's no StateDB set, it means we're not in an EthereumTx.
	if bk.StateDB == nil || !IsDeliverTx(ctx) {
		return
	}
...
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,7 +54,7 @@ See https://github.com/dangoslen/changelog-enforcer.
 - [#2348](https://github.com/NibiruChain/nibiru/pull/2348) - fix(oracle): max expiration a label rather than an invalidation for additional query liveness
 - [#2350](https://github.com/NibiruChain/nibiru/pull/2350) - fix(simapp): sim tests with empty validator set panic
 - [#2352](https://github.com/NibiruChain/nibiru/pull/2352) - chore(token-registry): Add bank coin versions of USDC and USDT from Stargate and LayerZero, and update ErisEvm.sol to fix redeem
-- [#2354](https://github.com/NibiruChain/nibiru/pull/2354) - chore: linter upgrade to v2
+- [#2357](https://github.com/NibiruChain/nibiru/pull/2357) - fix: proper statedb isolation in nibiru bank_extension
 
 ### Dependencies
 - Bump `form-data` from 4.0.1 to 4.0.4 ([#2347](https://github.com/NibiruChain/nibiru/pull/2347))


x/evm/keeper/keeper.go
```diff
@@ -122,3 +126,18 @@ func HandleOutOfGasPanic(err *error, format string) func() {
 		}
 	}
 }
+
+// IsSimulation checks if the context is a simulation context.
+func IsSimulation(ctx sdk.Context) bool {
```
k-yang, line 131: Is this still needed?

onikonychev: yes, here:
```go
// IsDeliverTx checks if we're in DeliverTx, NOT in CheckTx, ReCheckTx, or simulation
func IsDeliverTx(ctx sdk.Context) bool {
	return !ctx.IsCheckTx() && !ctx.IsReCheckTx() && !IsSimulation(ctx)
}
```

PR #2415, opened 2025-10-27
feat: custom ante NewDeductFeeDecorator allowing 0 fee for zero gas actors

onikonychev:
- copied the logic of https://github.com/cosmos/cosmos-sdk/blob/release/v0.47.x/x/auth/ante/fee.go
- customized it for the case when TX is sent by a zero gas actor

Key difference:
```go
// isZeroGasMeter checks if the context has a zero gas meter set.
// ZeroGas is detected in fixed_gas.go and sets NewFixedGasMeter(ZeroTxGas) to ctx
func isZeroGasMeter(ctx sdk.Context) bool {
	gasMeter := ctx.GasMeter()
	if _, ok := gasMeter.(*fixedGasMeter); ok {
		if gasMeter.GasConsumed() == ZeroTxGas && gasMeter.Limit() == math.MaxUint64 {
			return true
		}
	}
	return false
}
```

```go
if !isZeroGasMeter(ctx) {
	fee := feeTx.GetFee()
	if !simulate {
		fee, priority, err = dfd.txFeeChecker(ctx, tx)
		if err != nil {
			return ctx, err
		}
	}
	if err := dfd.checkDeductFee(ctx, tx, fee); err != nil {
		return ctx, err
	}
}
```

### DIFF
--- a/app/ante/deduct_fee.go
+++ b/app/ante/deduct_fee.go
@@ -6,14 +6,16 @@ import (
 
 	sdkmath "cosmossdk.io/math"
 
+	sdkioerrors "cosmossdk.io/errors"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 	authante "github.com/cosmos/cosmos-sdk/x/auth/ante"
 	"github.com/cosmos/cosmos-sdk/x/auth/types"
 )
 
-// DeductFeeDecorator deducts fees from the fee payer. The fee payer is the fee granter (if specified) or first signer of the tx.
-// If the fee payer does not have the funds to pay for the fees, return an InsufficientFunds error.
+// DeductFeeDecorator deducts fees from the fee payer. The fee payer is the fee
+// granter (if specified) or first signer of the tx. If the fee payer does not
+// have the funds to pay for the fees, return an InsufficientFunds error.
 // Call next AnteHandler if fees successfully deducted.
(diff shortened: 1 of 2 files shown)


app/ante/validator_tx_fee.go
Unique-Divine, line 1: Small change: Since all of this code is private and only used in the other file, can you move it to fee.go for better organization.

app/ante/fee.go
```diff
@@ -0,0 +1,149 @@
...
+		if !simulate {
+			fee, priority, err = dfd.txFeeChecker(ctx, tx)
+			if err != nil {
+				return ctx, err
+			}
+		}
+		if err := dfd.checkDeductFee(ctx, tx, fee); err != nil {
+			return ctx, err
+		}
+	}
```
Unique-Divine: You can move up the check with `isZeroGasMeter` to simplify this even more. You don't need the gas limit check if it's a zero gas sender. And because gas limit checks were completed earlier in the ante chain, you can honestly do:

```
if isZeroGasMeter(ctx) {
  return ctx, nil 
}
```

and skip this ante handler entirely

app/ante/validator_tx_fee.go
onikonychev: done

app/ante/fee.go
onikonychev: I think we still need next after the current handler
```go
return next(newCtx, tx, simulate)
```

Unique-Divine: ed44df9f548c08724f3a6ff1e643bc51a81f8ac7

Unique-Divine (approved):

PR #2508, opened 2026-01-23
feat: migrate to tachyon

matthiasmatt:
## Abstract

- Closes #AAA or https://github.../issue/AAA


Unique-Divine (changes requested):
I'm going to adjust this one so it doesn't rely on private modules. Probably easier to adjust the source code of our, bft repo and just change the replace statement to edit the version. 

It'll end up as a really small pull request after that.

PR #2702, opened 2026-07-21
fix(evm): Add nonce-aware EVM mempool and proposal selection; bound pending txs per sender

matthiasmatt:
# Add nonce-aware EVM mempool and proposal selection

Prevent a single EVM sender from flooding proposals with unusable future-nonce transactions by bounding node-local admission, retaining only valid state nonce chains through recheck, and constructing proposals from executable transactions.

## Key Changes

1. Add an EVM-aware SDK mempool that indexes transactions by authenticated sender, nonce, and CometBFT transaction key; limits each sender to the 64 nonce values from the state nonce through state nonce plus 63; and rejects conflicting transactions for occupied nonce slots.
1. Clean stale and gap-dependent transactions during `CheckTxType_Recheck`, and add idempotent transaction-key removal so failed rechecks and delivered transactions cannot remain in the SDK mempool.
1. Add an EVM-aware `PrepareProposal` implementation that preserves original transaction bytes, selects each sender's executable state nonce chain, prioritizes executable sender heads, respects block byte and gas limits, and retains CometBFT order for non-EVM transactions.
1. Keep proposal validation consensus-safe: node-local mempool contents affect candidate selection only, while proposal processing and delivery continue to require exact executable nonces.
1. Add coverage for admission boundaries, nonce collisions, recheck behavior, keyed cleanup, sender snapshots, proposal ordering, capacity limits, and original-byte preservation.

## Block commit and nonce chain cleanup

The SDK mempool retains unexecuted nonce slots across `EndBlock`. The block lifecycle reconciles those slots with the updated state nonce:

1. `DeliverTx` removes each included transaction from the SDK mempool by its CometBFT transaction key and advances the sender's state nonce.
1. ABCI `Commit` publishes the updated application state.
1. CometBFT removes the block's transactions from its consensus mempool.
1. With mempool setting `recheck = true`, CometBFT sends `CheckTxType_Recheck` for every remaining transaction.
1. The EVM ante flow retains a transaction only when every nonce from the updated state nonce through that transaction's nonce has a live slot. A failed recheck removes the transaction from both the CometBFT and SDK mempools.

For example, suppose the state nonce is `10`, and the live slots are `10, 12, 13`. If a block executes nonce `10`, the state nonce becomes `11`, leaving slots `12, 13`. Both remaining transactions fail recheck because nonce `11` is missing, so the node removes the disconnected tail. If the original slots were `10, 11, 12, 13` and the block executed `10, 11`, slots `12, 13` would survive because they form a complete state nonce chain beginning at the updated state nonce `12`.

CometBFT 0.37 enables `recheck` by default. Nodes must keep it enabled for automatic gap tail cleanup; `PrepareProposal` still excludes disconnected tails when recheck is disabled, but those transactions remain stored locally.

## Appendix

No CometBFT fork or consensus-state migration is required. The per-sender index is node-local, and CometBFT 0.37 recheck removes rejected transactions from its own mempool while BaseApp removes the corresponding SDK mempool entries.

Validated with:

```text
go test ./evm/... ./app/... ./lib/cosmos-sdk/baseapp ./lib/cosmos-sdk/types/mempool -count=1
```


evm/evmstate/evm_state.go
```diff
@@ -48,6 +48,9 @@ type EvmState struct {
 	BlockTxIndex collections.ItemTransient[uint64]
 	// BlockBloom: Bloom filters.
 	BlockBloom collections.ItemTransient[[]byte]
+	// PendingTxCount tracks node-local CheckTx admissions per EVM sender. It is
+	// reset with the transient store on commit and is never consensus state.
+	PendingTxCount collections.MapTransient[gethcommon.Address, uint64]
```
Unique-Divine: Using a KV store, even if transient affects gas metering and can cause consensus failures. Better to use a normal, async-compatible map. 

The intention of this is in the right direction, but it'll give nodes consensus errors writing different state into a KV store, even if it's transient.

There's an ABCI method for being able to inspect the mempool that allows the proper filtering as part of ABCI++ if we bump the CometBFT dependency up.

Because the ctx.TransientStore()  wraps gaskv (all stores do this), every Get, Insert, or anything meters gas via TransientKVGasConfig .

evm/evmante/evmante_increment_sender_seq.go
```diff
@@ -62,17 +72,35 @@ func AnteStepIncrementNonce(
...
+		)
+		if pendingTxCount >= MaxPendingTxsPerSender {
+			return fmt.Errorf(
+				"too many pending transactions for sender; got %d, limit %d",
+				pendingTxCount, MaxPendingTxsPerSender,
+			)
+		}
+		k.EVMState().PendingTxCount.Insert(
+			sdb.RootCtx(), msgEthTx.FromAddr(), pendingTxCount+1,
+		)
```
Unique-Divine: The ante handler is also not supposed to touch sdb.RootCtx() as that bypasses the intentional zeroing of gas configs for the active ctx. Small discrepenecies there can cause consensus failures and has done so in the past.

evm/evmstate/evm_state.go
Unique-Divine: Fixed in later change

Unique-Divine (changes requested):

evm/evmante/evmante_increment_sender_seq.go
Unique-Divine: Removed/fixed

PR #2727, opened 2026-07-28
fix(store/cache): prevent stale reads from inter-block cache races

matthiasmatt:
# Demonstration: inter-block cache stale-read race behind the testnet-2 fullnode-2 consensus halt

> **Draft on purpose — two of the three new tests fail on `main`, and that is the point.** They are regression tests for a live consensus bug; they must stay red until `CommitKVStoreCache` is synchronized, then gate the fix.

## The incident

`nibiru-testnet-2-fullnode-2` halted at height 8,658,454 (2026-07-26 23:08 UTC):

```
CONSENSUS FAILURE!!!
precommit step; +2/3 prevoted for an invalid block: wrong Block.Header.LastResultsHash.
Expected EA00F2F4A5382DDAF699144FF9999ED93B79D8AEFFD841365ECF64DCEA5134BE,
got      62DF2E9A0950A4EFBC676ECDD58E0EBB7770F6DF49E8739B88CA4C2EFD779633
```

## Verified evidence chain

Every step below was reproduced independently from live nodes, not inferred:

1. **The divergence is one tx result.** Recomputing CometBFT 0.37's `LastResultsHash` (deterministic `ResponseDeliverTx` proto + RFC 6962 Merkle root) from each node's `/block_results?height=8658454` reproduces both hashes byte-for-byte. The only difference: tx #2 (a SAI perp `trigger_trade_batch`) returned `code=5` on fullnode-2 vs `code=0` on every healthy node.
2. **The failing read was a phantom.** The `code=5` error was `state key not found` for the perp contract's `initial_acc_fees` key (token index 3, trade index 6329). A raw historical query **against fullnode-2's own disk** returns the value at heights 8658453 and 8658454 (absent at 8658452), byte-identical to the canonical archive node. The node's execution contradicted its own committed IAVL state — only a non-deterministic in-memory layer between execution and disk can do that.
3. **The code allows it.** `CommitKVStoreCache.Get` ([cache.go#L100-L115](https://github.com/NibiruChain/nibiru/blob/main/lib/cosmos-sdk/store/cache/cache.go#L100-L115)) is a read-through with no synchronization: cache miss → read underlying store → `cache.Add(key, value)`, caching `nil` misses too. `Set`/`Delete` are equally unsynchronized. The inter-block cache is on by default (`inter-block-cache = true`), and gRPC `Service/Simulate` executes against `checkState`, whose multistore branches over the **same shared** `CommitKVStoreCache` instances the commit path writes through.
4. **The interleaving:** a Simulate read misses the cache and reads "not found" from the pre-commit store → the block commit writes the key through the cache and to disk → the suspended read resumes and caches its stale `nil` over the fresh value → every subsequent block execution on that node reads the poisoned `nil` while the value sits intact on disk.
5. **Known upstream:** [cosmos/cosmos-sdk#23891](https://github.com/cosmos/cosmos-sdk/issues/23891) (open) reports this exact race; another network resolved identical app-hash failures by disabling the inter-block cache.

Why comparing blocks found nothing: headers matched through 8,658,454 — the divergence lives in the ABCI results, first committed by block 8,658,455's `LastResultsHash`. `/block_results?height=8658454` is where it shows.

## What this PR adds

Three tests in `lib/cosmos-sdk/store/cache/cache_race_test.go`:

| Test | What it shows | On `main` |
|---|---|---|
| `TestCommitKVStoreCacheStaleNilPoisoning` | Deterministic reproduction of the incident interleaving: stale `nil` cached over a freshly committed value; next read returns "not found" for a key that is on disk | **FAIL** |
| `TestCommitKVStoreCacheStaleValueResurrection` | Mirror image over `Delete`: a deleted key keeps resolving from the poisoned cache | **FAIL** |
| `TestCommitKVStoreCacheConcurrentGetSet` | The production access pattern (Simulate reads racing commit write-through); `go test -race` flags the data race on the underlying IAVL store | **FAIL under `-race`** |

The deterministic tests use a gated store wrapper to suspend a reader in the exact preemption window between its underlying read and the cache `Add` — no sleeps, no flakiness, 0.00s each.

```console
$ go test ./lib/cosmos-sdk/store/cache/ -run 'StaleNil|Resurrection' -v
--- FAIL: TestCommitKVStoreCacheStaleNilPoisoning (0.00s)
        Messages:   phantom 'state key not found': cache returned stale nil for a key that is on disk
--- FAIL: TestCommitKVStoreCacheStaleValueResurrection (0.00s)
        Messages:   phantom resurrection: cache returned a value that was deleted on disk

$ go test -race ./lib/cosmos-sdk/store/cache/ -run ConcurrentGetSet
WARNING: DATA RACE
  Read:  iavl.(*MutableTree).Get  ← CommitKVStoreCache.Get  cache.go:111
  Write: iavl.(*MutableTree).set  ← CommitKVStoreCache.Set  cache.go:124
```

## The tests gate a real fix

Verified locally (not included in this PR — fix design is the team's call): with a naive mutex over `Get`/`Set`/`Delete`, **all three tests pass, including under `-race`**, and all pre-existing cache tests stay green:

```diff
 	CommitKVStoreCache struct {
 		types.CommitKVStore
 		cache *lru.ARCCache
+		mtx   sync.Mutex
 	}
```
plus `ckv.mtx.Lock(); defer ckv.mtx.Unlock()` at the top of `Get`, `Set`, and `Delete` (8 lines total). A big lock serializes underlying IAVL reads, so something finer-grained (e.g. skip caching when a write raced the read) may be preferable for throughput — the tests will hold either way.

## Interim mitigation

Until a fix lands: `inter-block-cache = false` in `app.toml` on every node serving asynchronous query/Simulate traffic. Note the recorded divergent ABCI result on fullnode-2 is persisted — a restart cannot repair it; roll back one height or re-sync.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/cosmos-sdk/store/cache/cache.go
+++ b/lib/cosmos-sdk/store/cache/cache.go
@@ -2,6 +2,7 @@ package cache
 
 import (
 	"fmt"
+	"sync"
 
 	"github.com/NibiruChain/nibiru/v2/lib/cosmos-sdk/store/cachekv"
 	"github.com/NibiruChain/nibiru/v2/lib/cosmos-sdk/store/types"
@@ -28,6 +29,7 @@ type (
 	CommitKVStoreCache struct {
 		types.CommitKVStore
 		cache *lru.ARCCache
+		mtx   sync.RWMutex
 	}
 
 	// CommitKVStoreCacheManager maintains a mapping from a StoreKey to a
@@ -100,6 +102,9 @@ func (ckv *CommitKVStoreCache) CacheWrap() types.CacheWrap {
 func (ckv *CommitKVStoreCache) Get(key []byte) []byte {
 	types.AssertValidKey(key)
 
+	ckv.mtx.RLock()
+	defer ckv.mtx.RUnlock()
+
 	keyStr := string(key)
 	valueI, ok := ckv.cache.Get(keyStr)
 	if ok {
@@ -120,6 +125,9 @@ func (ckv *CommitKVStoreCache) Set(key, value []byte) {
 	types.AssertValidKey(key)
 	types.AssertValidValue(value)
 
+	ckv.mtx.Lock()
+	defer ckv.mtx.Unlock()
+
 	ckv.cache.Add(string(key), value)
 	ckv.CommitKVStore.Set(key, value)
 }
@@ -127,6 +135,9 @@ func (ckv *CommitKVStoreCache) Set(key, value []byte) {
 // Delete removes a key/value pair from both the write-through cache and the
 // underlying CommitKVStore.
 func (ckv *CommitKVStoreCache) Delete(key []byte) {
+	ckv.mtx.Lock()
+	defer ckv.mtx.Unlock()
+
 	ckv.cache.Remove(string(key))
 	ckv.CommitKVStore.Delete(key)
 }
--- a/lib/cosmos-sdk/store/cache/cache_race_test.go
+++ b/lib/cosmos-sdk/store/cache/cache_race_test.go
@@ -1,0 +1,211 @@
+package cache_test
+
+// Regression tests for the inter-block cache race that halted the
+// nibiru-testnet-2-fullnode-2 node at height 8658454 with
+// "CONSENSUS FAILURE!!! ... wrong Block.Header.LastResultsHash".
+//
+// CommitKVStoreCache.Get is a read-through: on a cache miss it reads the
+// underlying CommitKVStore and then caches the result (including nil misses)
+// with no synchronization against concurrent Set/Delete. gRPC Service/Simulate
+// traffic executes against checkState, whose CacheMultiStore branches over the
+// same shared CommitKVStoreCache instances that block commit writes through.
+// A simulation read racing a commit can therefore overwrite a freshly
+// committed value with a stale nil; every subsequent block execution on that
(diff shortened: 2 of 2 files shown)


Unique-Divine:
I implemented a targeted synchronization fix locally on this PR branch (commit `e65bad7b2`; not pushed yet).

The change adds one `sync.RWMutex` to each `CommitKVStoreCache`:

- Method `Get` holds a read lock across the ARC lookup, underlying store read, and cache fill.
- Methods `Set` and `Delete` hold an exclusive lock across both the ARC mutation and underlying store mutation.

The lock must cover the complete cache-miss sequence. Otherwise, a reader can observe a missing key, pause, and then cache that stale `nil` after a concurrent commit has written the value. This is the interleaving demonstrated by `TestCommitKVStoreCacheStaleNilPoisoning`. An `RWMutex` also allows concurrent reads while preventing reads from racing IAVL writes.

The three regression tests pass with the change, including the focused race-detector run:

```console
go test -race ./store/cache -count=1
```

This is a narrow fix for the demonstrated inter-block cache poisoning. It does not provide a versioned snapshot for an entire simulation, so the broader simulation-versus-commit consistency design described in cosmos/cosmos-sdk#20685 remains separate work. CI can run the complete Cosmos SDK suite.

Unique-Divine (approved):
Works with small, defensive locks

